PR #6, opened 2019-06-20
fix centerline nan issue

alliecc:
Filter out nan values in center line points


demo_usage/cuboids_to_bboxes.py
```diff
@@ -77,6 +77,10 @@ def plot_lane_centerlines_in_img(
 
         ground_heights = avm.get_ground_height_at_xy(centerline_city_fr, city_name)
 
+        valid_idx = np.isnan(ground_heights)
+        centerline_city_fr = centerline_city_fr[~valid_idx]
+
+
```
mgilson-argo: nit: only one blank line.

PR #13, opened 2019-06-20
Loop over cameras (not lidar sweeps) when rendering

johnwlambert:
Avoid the optical illusion from plotting annotations from different sweeps in the same image. We switch to looping over images, not PLY fpaths. Fixed typo in unit test.

Rendering now is at 30 fps, not 10 fps (see [this link](https://www.youtube.com/watch?v=iOxbLH8Aq9Q&feature=youtu.be) for an example of new functionality provided by this PR.

### DIFF
--- a/argoverse/data_loading/simple_track_dataloader.py
+++ b/argoverse/data_loading/simple_track_dataloader.py
@@ -92,6 +92,12 @@ class SimpleArgoverseTrackingDataLoader:
 
     def get_closest_lidar_fpath(self, log_id: str, cam_timestamp: int) -> Optional[str]:
         """
+        Args:
+            log_id: str, unique ID of vehicle log
+            cam_timestamp: int, timestamp of image capture, in nanoseconds
+
+        Returns:
+            ply_fpath: str, string representing path to PLY file, or else None.
         """
         lidar_timestamp = self.sdb.get_closest_lidar_timestamp(cam_timestamp, log_id)
         if lidar_timestamp is None:
@@ -114,8 +120,11 @@ class SimpleArgoverseTrackingDataLoader:
     def get_ordered_log_cam_fpaths(self, log_id: str, camera_name: str) -> List[str]:
         """
         Args
+            log_id: str, unique ID of vehicle log
 
         Returns
+            cam_img_fpaths: List of strings, representing paths to JPEG files in this log,
+                for a specific camera
         """
         cam_img_fpaths = sorted(glob.glob(f"{self.data_dir}/{log_id}/{camera_name}/{camera_name}_*.jpg"))
         return cam_img_fpaths
--- a/argoverse/data_loading/simple_track_dataloader.py
+++ b/argoverse/data_loading/simple_track_dataloader.py
@@ -54,16 +54,16 @@ class SimpleArgoverseTrackingDataLoader:
         assert isinstance(log_calib_data, dict)
         return log_calib_data
 
-    def get_city_to_egovehicle_se3(self, log_id: str, lidar_timestamp: int) -> Optional[SE3]:
+    def get_city_to_egovehicle_se3(self, log_id: str, timestamp: int) -> Optional[SE3]:
         """
         Args:
             log_id: str, unique ID of vehicle log
-            lidar_timestamp: int, timestamp of LiDAR sweep capture, in nanoseconds
+            timestamp: int, timestamp of sensor observation, in nanoseconds
 
         Returns:
             city_to_egovehicle_se3: SE3 transformation to bring egovehicle frame point into city frame.
         """
-        pose_fpath = f"{self.data_dir}/{log_id}/poses/city_SE3_egovehicle_{lidar_timestamp}.json"
+        pose_fpath = f"{self.data_dir}/{log_id}/poses/city_SE3_egovehicle_{timestamp}.json"
         if not Path(pose_fpath).exists():
             return None
         pose_data = read_json_file(pose_fpath)
@@ -90,6 +90,17 @@ class SimpleArgoverseTrackingDataLoader:
         im_fpath = f"{im_dir}/{im_fname}"
         return im_fpath
 
+    def get_closest_lidar_fpath(self, log_id: str, cam_timestamp: int) -> Optional[str]:
+        """
+        """
+        lidar_timestamp = self.sdb.get_closest_lidar_timestamp(cam_timestamp, log_id)
+        if lidar_timestamp is None:
+            return None
+        lidar_dir = f"{self.data_dir}/{log_id}/lidar"
+        ply_fname = f"PC_{lidar_timestamp}.ply"
+        ply_fpath = f"{lidar_dir}/{ply_fname}"
+        return ply_fpath
+
     def get_ordered_log_ply_fpaths(self, log_id: str) -> List[str]:
         """
         Args:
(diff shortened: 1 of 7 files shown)


argoverse/data_loading/simple_track_dataloader.py
```diff
@@ -54,16 +54,16 @@ def get_log_calibration_data(self, log_id: str) -> Mapping[str, Any]:
         assert isinstance(log_calib_data, dict)
         return log_calib_data
 
-    def get_city_to_egovehicle_se3(self, log_id: str, lidar_timestamp: int) -> Optional[SE3]:
+    def get_city_to_egovehicle_se3(self, log_id: str, timestamp: int) -> Optional[SE3]:
```
johnwlambert: we can get city_SE3_egovehicle at camera or LiDAR observation time, so this should be agnostic. I made it more generic in the name.

mriley-argo:
@johnwlambert Can you rebase your changes on the latest `master`?  The test failure here has been fixed there, and hopefully they'll pass on this branch.

johnwlambert:
@mgilson-argo @mriley-argo Looks like Travis passed, so should be ready for a merge now.

argoverse/data_loading/simple_track_dataloader.py
```diff
@@ -100,6 +117,18 @@ def get_ordered_log_ply_fpaths(self, log_id: str) -> List[str]:
...
+    def get_ordered_log_cam_fpaths(self, log_id: str, camera_name: str) -> List[str]:
+        """
+        Args
+            log_id: str, unique ID of vehicle log
+
+        Returns
+            cam_img_fpaths: List of strings, representing paths to JPEG files in this log,
+                for a specific camera
+        """
+        cam_img_fpaths = sorted(glob.glob(f"{self.data_dir}/{log_id}/{camera_name}/{camera_name}_*.jpg"))
```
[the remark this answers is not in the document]
johnwlambert: @mgilson-argo The image files are zero-padded to nanosecond values, so we don't have an issue with lexicographic sort in such a case.

johnwlambert:
@mgilson-argo Good point about adding more unit tests for the new functionality, I will do this later -- let's hold off on the merge then.
-John

johnwlambert:
@mgilson-argo Hey Matt, please review, should be ready for merge

demo_usage/cuboids_to_bboxes.py
```diff
@@ -122,37 +121,41 @@ def dump_clipped_3d_cuboids_to_images(
...
-            for cam_idx, camera_name in enumerate(RING_CAMERA_LIST + STEREO_CAMERA_LIST):
-                camera_counter = 0
-                # load images, e.g. 'image_raw_ring_front_center_000000486.jpg'
-                im_fpath = dl.get_closest_im_fpath(log_id, camera_name, lidar_timestamp)
-                if im_fpath is None:
+        for cam_idx, camera_name in enumerate(RING_CAMERA_LIST + STEREO_CAMERA_LIST):
+            cam_im_fpaths = dl.get_ordered_log_cam_fpaths(log_id, camera_name)
+            for i, im_fpath in enumerate(cam_im_fpaths):
+                if i % 50 == 0:
+                    logging.info(f"\tOn file {i} of camera {camera_name} of {log_id}")
```
mgilson-argo: We should let the logger handle the string formatting:

```py
logging.info("\tOn file %s of camera %s of %s", i, camera_name, log_id)
```

demo_usage/cuboids_to_bboxes.py
```diff
@@ -122,37 +121,41 @@ def dump_clipped_3d_cuboids_to_images(
...
+
+                cam_timestamp = im_fpath.split("/")[-1].split(".")[0].split("_")[-1]
```
mgilson-argo: Can we use `pathlib` or `os.path` to do this manipulation instead of string splitting?

demo_usage/cuboids_to_bboxes.py
```diff
@@ -122,37 +121,41 @@ def dump_clipped_3d_cuboids_to_images(
...
                     if max_num_images_to_render != -1 and len(saved_img_fpaths) > max_num_images_to_render:
                         flag_done = True
                         break
                     continue
+
+                city_to_egovehicle_se3 = dl.get_city_to_egovehicle_se3(log_id, cam_timestamp)
+                if city_to_egovehicle_se3 is None:
+                    continue
+
+                lidar_timestamp = ply_fpath.split("/")[-1].split(".")[0].split("_")[-1]
```
mgilson-argo: Same comment here.

demo_usage/cuboids_to_bboxes.py
```diff
@@ -122,37 +121,41 @@ def dump_clipped_3d_cuboids_to_images(
...
+                lidar_timestamp = int(lidar_timestamp)
+                labels = dl.get_labels_at_lidar_timestamp(log_id, lidar_timestamp)
+                if labels is None:
+                    logger.info(f"\tLabels missing at t={lidar_timestamp}")
```
mgilson-argo: Same comment about the logging.

johnwlambert: Thanks, what's the advantage @mgilson-argo ?

demo_usage/cuboids_to_bboxes.py
johnwlambert: Should that be `logging.info` or `logger.info`? we're using both

johnwlambert:
@mgilson-argo Swapped in pathlib

demo_usage/cuboids_to_bboxes.py
mgilson-argo: If someone set up a module-level logger, then I'd use that.  `logging.info` will use the root logger.  `logger.info` will use the module specific logger (assuming that `logger` was created by calling `logging.getLogger(some_name)`)

demo_usage/cuboids_to_bboxes.py
mgilson-argo: First and foremost, it's considered a "best-practice".  The reason that it's a best practice can be demonstrated by the following class:

```py
def Foo:
    def __str__(self) -> str:
        time.sleep(10)
        return "Well, that took a while..."
```

If you log this thing:

```py
instance = Foo()
logging.debug(f"instance: {instance}")
```

You'll take ~10s to make the log statement even if your log level is set such that you'll never get a chance to see the message.  If you defer the formatting by passing it as format strings to the logger:

```py
logging.debug("instance: %s", instance)
```

then you'll save yourself 10s each time you hit this line (unless your log level is `DEBUG` or lower).  Basically, it's a performance boost for formatting expensive strings.  Most of the time it doesn't matter, but once in a while it does.  Because of that, it's a best practice to just always defer the formatting and let the logger handle it.

(the logger will also do things like catch and re-log errors if you happen to specify the format string incorrectly whereas doing it greedily will cause exceptions -- though this is less of a problem in a world where we have `f-strings`).

argoverse/data_loading/synchronization_database.py
```diff
@@ -62,23 +90,50 @@ def __init__(self, dataset_dir: str, collect_single_log_id: Optional[str] = None
...
+            self.per_log_lidartimestamps_index[log_id] = lidar_timestamps
 
     def get_valid_logs(self) -> Iterable[str]:
         """ Return the log_ids for which the SynchronizationDatabase contains pose information.
         """
-        return self.per_log_pose_index.keys()
+        return self.per_log_camtimestamps_index.keys()
+
+    def get_closest_lidar_timestamp(self, cam_timestamp: int, log_id: str) -> Optional[int]:
+        """
```
mgilson-argo: Docstring?

johnwlambert: @mgilson-argo thanks for catching that, just added it

mgilson-argo:
Looks OK to me.  I'll defer to @ivolkwein on merging.

PR #148, opened 2020-07-10
Add the python file for computing difficulty attributes

alliecc:


argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
+import sys
```
johnwlambert, line 1: @alliecc thanks for adding this, Allie! I think the `sys` import is unused, can we remove it? Also can we order the imports in the following order:

1. Python builtins
(space)
2. external libraries
(space)
3. then argoverse modules?

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
...
+import glob
+import scipy
+import scipy.interpolate as interpolate
+import torch.nn.functional as F
+import torch
+import cv2
+import pickle
+
+dict_color = {}
+dict_color["easy"] = (0, 1.0, 0)
```
johnwlambert: could we add a small comment next to these, with their qualitative color description, e.g. light-green, etc?

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
...
+
+font = cv2.FONT_HERSHEY_SIMPLEX
+fontScale = 1.2
+fontColor = (255, 255, 255)
+lineType = 3
+visualize = True
+check_track_label_folder = True
+
+
+def save_bev_img(path_output_vis, list_bboxes, list_difficulty_att,
```
johnwlambert: could we add type hints here as well?

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
...
+        np.logical_and(pc[:, 0] < image_size, pc[:, 1] < image_size))
+    img[pc[ind_valid, 0], pc[ind_valid, 1], :] = 0.4
+
+    path_imgs = os.path.join(path_output_vis, 'bev')
+    if not os.path.exists(path_imgs):
+        os.mkdir(path_imgs)
+
+    for bbox, difficulty_att in zip(list_bboxes, list_difficulty_att):
+
+        q0, q1, q2, q3 = bbox['rotation']['w'], bbox['rotation']['x'], bbox[
```
johnwlambert, line 60: can we name these `qw, qx, qy, qz` instead of `q0, q1, q2, q3`, since different libraries order those quaternion elements differently, but w,x,y,z names are always respected?

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
...
+        offset += 150
+
+    print('Saving img: ', path_imgs)
+    cv2.imwrite(
+        os.path.join(path_imgs,
+                     '%s_%s_%s.jpg' % (dataset_name, log_id, lidar_timestamp)),
+        img * 255)
+
+
+def bspline_1d(x, y, s=20, k=3):
```
johnwlambert: type hint here as well :-)

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
...
+        return y
+
+    tck = interpolate.splrep(x, y[x], s=s, k=k)
+
+    return interpolate.splev(np.arange(y.shape[0]), tck)
+
+
+def derivative(x):
+    """
+    Compute derivative for velocity and acceleration
```
johnwlambert, line 129: could you add a small comment here with the input and output array/tensor sizes?

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
...
+        (vx[:, np.newaxis], vy[:, np.newaxis], vz[:, np.newaxis]), axis=1)
+    a = np.concatenate(
+        (ax[:, np.newaxis], ay[:, np.newaxis], az[:, np.newaxis]), axis=1)
+
+    v[np.isnan(traj[:, 0])] = np.nan
+    a[np.isnan(traj[:, 0])] = np.nan
+    return v, a
+
+
+def read_json_file(fpath):
```
johnwlambert: can we use the library instead of re-defining this helper function? it's provided in 
`from argoverse.utils.json_utils import read_json_file`

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
...
+    with open(fpath, "rb") as f:
+        return json.load(f)
+
+
+def save_json_file(fpath, x):
```
johnwlambert: same here for `from argoverse.utils.json_utils import save_json_dict` :-)

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
...
+    with open(fpath, "w") as f:
+        return json.dump(x, f)
+
+
+# set root_dir to the correct path to your dataset folder
+root_dir = 'test_set/'
+path_output_vis = 'vis_output'
+filename_output = 'att_file.npy'
+
+if not os.path.exists(path_output_vis):
```
johnwlambert, line 190: can we put these in a separate function that is called within `if __name__ == __main__`, so that if the file is imported as a module, they aren't automatically run?

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
...
+            data = data_amodel[id_track]
+            if data["label_class"] not in list_name_class:
+                continue
+
+            data_per_frame = data['track_label_frames']
+            dict_tracks[id_track] = {}
+            dict_tracks[id_track]["ind_lidar_min"] = -1
+            dict_tracks[id_track]["ind_lidar_max"] = -1
+            length_log = len(list_lidar_timestamp)
+            dict_tracks[id_track]["list_world_se3"] = [None] * length_log
```
johnwlambert, line 249: can we rename `world` as `city`, just to be consistent with the paper/elsewhere in codebase?

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,380 @@
...
+
+        for id_track in dict_tracks.keys():
+            list_key = list(dict_tracks[id_track].keys()).copy()
+            for key in list_key:
+                if key != "difficult_att":
+                    del dict_tracks[id_track][key]
+
+        dict_att_all[name_folder][id_log] = dict_tracks
+
+pickle_out = open(filename_output, "wb")
```
johnwlambert, line 380: we have this function in the api too, can we call it as a library function?

`from argoverse.utils.pkl_utils import save_pkl_dictionary`

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+    """
+    image_size = 2000
+    image_scale = 10
+    img = np.zeros((image_size, image_size, 3))
+    pc = pc * image_scale
+    pc[:, 0] += int(image_size / 2)
+    pc[:, 1] += int(image_size / 2)
+    pc = pc.astype("int")
+
+    ind_valid = np.logical_and(
```
johnwlambert, line 57: you can actually do this with `np.logical_and.reduce()` and just pass in a list of predicates, instead of chaining several `np.logical_and()` calls together.

>>> `import numpy as np`
>>> `x = np.array([1,2,3,4,5])`
>>> `np.logical_and.reduce( [x>=2, x <=4] )`
array([False,  True,  True,  True, False], dtype=bool)
>>>

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+    if len(ind_valid) > 0:
+        x_max = dx[ind_valid].max()
+        x_min = dx[ind_valid].min()
+        y_max = dy[ind_valid].max()
+        y_min = dy[ind_valid].min()
+
+        if np.sqrt((x_max - x_min) ** 2 + (y_max - y_min) ** 2) < 10:
+            return np.zeros_like(traj), np.zeros_like(traj)
+
+    vx = derivative(dx) * 10  # lidar freq 10 hz
```
johnwlambert, line 164: we could make this a private module-level constant, like
`LIDAR_FPS=10 # 10 Hz frequency` to prevent the hard-coded `10`s

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+                    if egovehicle_to_city_se3 is None:
+                        print("Pose not found!")
+                        continue
+                    center_w = egovehicle_to_city_se3.transform_point_cloud(center[np.newaxis, :])[0]
+
+                    dict_tracks[id_track]["list_center"][ind_lidar] = center
+                    dict_tracks[id_track]["list_center_w"][ind_lidar] = center_w
+                    dict_tracks[id_track]["list_dist"][ind_lidar] = np.linalg.norm(center[0:2])
+                    dict_tracks[id_track]["exists"][ind_lidar] = True
+                    dict_tracks[id_track]["list_city_se3"][ind_lidar] = egovehicle_to_city_se3
```
johnwlambert, line 279: I usually used the notation `city_SE3_egovehicle` instead of `egovehicle_to_city_se3` just so that the semantics are
`p_city = city_SE3_egovehicle * p_egovehicle`. You are expressing the transformation from the egovehicle frame, to the city, right? I think something similar is often [done in GTSAM too](https://gtsam.org/gtsam.org/2020/06/28/gtsam-conventions.html)

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+                dist_close = 50
+                ind_valid = np.nonzero(1 - np.isnan(dict_tracks[id_track]["list_dist"]))[0]
+                ind_close = np.nonzero(dict_tracks[id_track]["list_dist"][ind_valid] < dist_close)[0]
+
+                if len(ind_close) > 0:
+                    ind_close_max = ind_close.max() + 1
+                    ind_close_min = ind_close.min()
+
+                if dict_tracks[id_track]["length_track"] == 0:
+                    # this shouldn't happen
```
johnwlambert, line 308: should we `raise Exception` here or `raise AssertionError` to catch this potential error?

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+                    list_difficulty_att = []
+
+                    for id_track in dict_tracks.keys():
+                        if dict_tracks[id_track]["exists"][ind_lidar]:
+                            list_bboxes.append(dict_tracks[id_track]["list_bbox"][ind_lidar])
+                            list_difficulty_att.append(dict_tracks[id_track]["difficult_att"])
+
+                    path_lidar = os.path.join(path_log, "lidar", "PC_%s.ply" % timestamp_lidar)
+                    egovehicle_to_city_se3 = argoverse_loader.get_pose(ind_lidar, id_log)
+                    pc = np.asarray(o3d.io.read_point_cloud(path_lidar).points)
```
johnwlambert, line 342: does open3d support reading the reflectance values now? We could deprecate pyntcloud and switch to open3d altogether if so

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+            color_1 += dict_color[att][1] / len(difficulty_att)
+            color_2 += dict_color[att][2] / len(difficulty_att)
+        color = (color_0, color_1, color_2)
+
+        w, l, h = bbox["width"], bbox["length"], bbox["height"]
+        bbox_2d = np.array([[-l / 2, -w / 2, 0], [l / 2, -w / 2, 0], [-l / 2, w / 2, 0], [l / 2, w / 2, 0]])
+        R = np.array(
+            [[np.cos(theta_local), -np.sin(theta_local), 0], [np.sin(theta_local), np.cos(theta_local), 0], [0, 0, 1]]
+        )
+        bbox_2d = np.matmul(R, bbox_2d.transpose()).transpose() + pose_local[0:3]
```
johnwlambert, line 86: For the sake of simplicity, might be easier just to use the following snippet:
```
egovehicle_SE3_obj = SE3(rotation=quat2rotmat(self.quaternion), translation=self.translation)
bbox_ego_frame = egovehicle_SE3_obj .transform_point_cloud(bbox_object_frame)
```
Or alternatively we could just use the `as_2d_bbox()` function [here](https://github.com/argoai/argoverse-api/blob/master/argoverse/data_loading/object_label_record.py#L53). 
after `json_label_dict_to_obj_record(label)`

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+    offset = 0
+    for key in dict_color.keys():
+        cv2.putText(img, key, (100 + offset, image_size - 50), font, fontScale, dict_color[key], lineType)
+        offset += 150
+
+    print("Saving img: ", path_imgs)
+    cv2.imwrite(os.path.join(path_imgs, "%s_%s_%d.jpg" % (dataset_name, log_id, lidar_timestamp)), img * 255)
+
+
+def bspline_1d(x: np.array, y: np.array, s: float = 20.0, k: int = 3) -> np.array:
```
johnwlambert, line 121: can we add the shape of `x`, `y`? And can we add what `s` and `k` represent in the B-Spline computation?

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+    """
+    Do Bspline smoothing
+    """
+
+    if len(x) <= k:
+        return y
+
+    tck = interpolate.splrep(x, y[x], s=s, k=k)
```
johnwlambert, line 129: are we interpolating a trajectory? maybe `traj` is more appropriate than `tck`? or `track` might be more explicit than `tck`

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+    v = np.concatenate((vx[:, np.newaxis], vy[:, np.newaxis], vz[:, np.newaxis]), axis=1)
+    a = np.concatenate((ax[:, np.newaxis], ay[:, np.newaxis], az[:, np.newaxis]), axis=1)
+
+    v[np.isnan(traj[:, 0])] = np.nan
+    a[np.isnan(traj[:, 0])] = np.nan
+    return v, a
+
+
+if __name__ == "__main__":
+    # set root_dir to the correct path to your dataset folder
```
johnwlambert, line 181: maybe we could make this long function a separate function named `make_att_files()`?

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+    list_name_class = ["VEHICLE", "PEDESTRIAN"]
+    count_track = 0
+    dict_att_all: Dict[str, Any] = {}
+    for name_folder in list_folders:
+
+        dict_att_all[name_folder] = {}
+        list_log_folders = glob.glob(os.path.join(root_dir, name_folder, "*"))
+        for ind_log, path_log in enumerate(list_log_folders):
+
+            id_log = path_log.split("/")[-1]
```
johnwlambert, line 199: I would probably prefer
```python
id_log = f'{Path(path_log).name}' to avoid the '/' splitting
```

argoverse/utils/make_att_files.py
```diff
@@ -0,0 +1,365 @@
...
+                list_path_label_persweep.sort()
+
+                dict_track_labels: Dict[str, Any] = {}
+                for path_label_persweep in list_path_label_persweep:
+                    data = read_json_file(path_label_persweep)
+                    for data_obj in data:
+                        id_obj = data_obj["track_label_uuid"]
+
+                        if id_obj not in dict_track_labels.keys():
+                            dict_track_labels[id_obj] = []
```
johnwlambert, line 213: we can avoid the `if` branch in two ways :-) 
either use `collections.defaultdict`, or use `dict_track_labels.get(id_obj, [])`

argoverse/utils/make_att_files.py
alliecc: Maybe we could use the xyzirgb format here? http://www.open3d.org/docs/release/tutorial/Basic/file_io.html#Point-cloud

argoverse/utils/make_att_files.py
alliecc: I think tck is more representative for bspline interpolation. I would prefer tck here since it's inside the bspline function.

argoverse/utils/make_att_files.py
alliecc: I think it's slightly different here. It's making an empty list for adding items in the future, not returning empty list when the key doesn't exist.

argoverse/utils/make_att_files.py
alliecc: I am not sure if using an SE object would increase the simplicity here.

argoverse/utils/make_att_files.py
johnwlambert: oh nevermind .get() will only work for retrieving. For setting those keys 

```python
dict_track_labels.setdefault(id_obj, [])
```
should work

johnwlambert: "This method returns the key value available in the dictionary and if given key is not available then it will return provided default value."
https://www.tutorialspoint.com/python/dictionary_setdefault.htm

PR #149, opened 2020-07-11
Refine make_att_files.py

alliecc:


argoverse/utils/make_att_files.py
```diff
@@ -304,27 +306,22 @@ def compute_v_a(traj: np.array) -> Tuple[np.array, np.array]:  # traj:Nx3
...
-                        if dict_tracks[id_track]["length_track"] < 10 or dict_tracks[id_track]["exists"].sum() < 10:
-                            dict_tracks[id_track]["difficult_att"].append("short")
-                        else:
-                            if (ind_close_max - ind_close_min) - dict_tracks[id_track]["exists"][
-                                ind_close_min:ind_close_max
-                            ].sum() > 20:
-                                dict_tracks[id_track]["difficult_att"].append("occ")
+                        if (ind_close_max - ind_close_min) - dict_tracks[id_track]["exists"][
+                            ind_close_min:ind_close_max
+                        ].sum() > 20:
```
johnwlambert, line 318: can we make 20 a module-level private variable? e.g. `MAX_OCCLUSION_PCT_ = 20`

argoverse/utils/make_att_files.py
```diff
@@ -304,27 +306,22 @@ def compute_v_a(traj: np.array) -> Tuple[np.array, np.array]:  # traj:Nx3
...
-                if dict_tracks[id_track]["length_track"] == 0:
-                    # this shouldn't happen
-                    dict_tracks[id_track]["difficult_att"].append("zero_length")
-
+                if dict_tracks[id_track]["list_dist"][ind_valid].min() > dist_close:
+                    dict_tracks[id_track]["difficult_att"].append("far")
                 else:
-                    if dict_tracks[id_track]["list_dist"][ind_valid].min() > dist_close:
-                        dict_tracks[id_track]["difficult_att"].append("far")
+                    if dict_tracks[id_track]["length_track"] < 10 or dict_tracks[id_track]["exists"].sum() < 10:
```
johnwlambert, line 313: can we make 10 a module-level private variable? e.g. `NEAR_DISTANCE_THRESH_ = 10`

argoverse/utils/make_att_files.py
```diff
@@ -161,13 +163,13 @@ def compute_v_a(traj: np.array) -> Tuple[np.array, np.array]:  # traj:Nx3
         if np.sqrt((x_max - x_min) ** 2 + (y_max - y_min) ** 2) < 10:
```
johnwlambert, line 163: could we add a note about why a B-spline is needed in the function comment? I guess for occluded frames to get a constant tick rate (exactly 100 ms)?

argoverse/utils/make_att_files.py
alliecc: It's not a distance threshold but a time threshold. I added a variable for it.

argoverse/utils/make_att_files.py
alliecc: It's for general temporal smoothing. Not only for the occluded frames. I will add the comment.

argoverse/utils/make_att_files.py
alliecc: ok

PR #153, opened 2020-07-19
remove filter and add length check for camera/lidar syncronization

janesjanes:


argoverse/data_loading/argoverse_tracking_loader.py
```diff
@@ -171,16 +171,15 @@ def image_list_sync(self) -> Dict[str, List[np.ndarray]]:
...
-                                    self.sync.get_closest_cam_channel_timestamp(x, camera, log)
-                                    for x in self._lidar_timestamp_list[log]
-                                ),
-                            )
+                            self.sync.get_closest_cam_channel_timestamp(x, camera, log)
+                            for x in self._lidar_timestamp_list[log]
                         ),
                     )
 
+                    assert len(self._image_timestamp_list_sync[log][camera]) == len(
```
johnwlambert, line 179: Thanks for this PR, Patsorn. Can we check on which logs from train1/train2/train3/val/test this assert passes? Another option would be to make an argument to the constructor that states in this rare edge case I prefer "None" or the closest image forwards or backwards, or the closest stale image (backwards only)

janesjanes: That assertion always pass, even when we have a missing image (we replace the missing image with None or closest image), it's just a sanity check :) 
I added the option to choose what to return when there are missing image, current version of argoverse dataset doesn't have anything missing in the ring camera, though there are some stereo image timestamp missing, which hopefully this option would be useful.

argoverse/data_loading/argoverse_tracking_loader.py
```diff
@@ -459,10 +480,14 @@ def get_image_at_timestamp(
         if log_id is None:
             log_id = self.current_log
         assert self.timestamp_image_dict is not None
+
+        if timestamp is None:
+            return None
+
```
johnwlambert, line 486: was there a specific reason we need this check? may a better one would be
```python
assert isinstance(timestamp, int), "query timestamp provided was not an integer"
```
since timestamp should never be none?

argoverse/data_loading/argoverse_tracking_loader.py
```diff
@@ -36,7 +36,14 @@ def _read_city_name(path: str) -> str:
 
 
 class ArgoverseTrackingLoader:
-    def __init__(self, root_dir: str) -> None:
+    def __init__(self, root_dir: str, return_missing_image_as: Optional[str] = None) -> None:
+        """
+        root_dir: directory of the dataset
+        return_missing_image_as: specify what to return when there are image with missing timestamp (default = None).
+                            None        -> None,
+                            'backward'  -> closest previous timestamp,
```
johnwlambert, line 44: this looks good. Is there a specific use case when you think the next or previous image would be preferred?

johnwlambert (changes requested):

argoverse/data_loading/argoverse_tracking_loader.py
janesjanes: I guess I'm using None timestamp to represent missing image, so it could happen in this case. (Line 206 for example) We could check it before calling the function, but I don't see much of a different either way?

argoverse/data_loading/argoverse_tracking_loader.py
janesjanes: Hmm not sure what'd be the use case, just feel like if we have backward only we should probably have forward only as well, in case they want that.
But... I forgot the closest in either direction case, haha. gonna add that

johnwlambert:
@janesjanes looks good, do you mind adding 3 simple unit tests with toy data to make sure the forward, backward, and closest work as expected?

argoverse/data_loading/argoverse_tracking_loader.py
```diff
@@ -171,20 +181,66 @@ def image_list_sync(self) -> Dict[str, List[np.ndarray]]:
...
+                                for ii in range(
+                                    i + 1, len(self._image_timestamp_list_sync[log][camera])
+                                ):  # loop forward
+                                    x = self._image_timestamp_list_sync[log][camera][ii]
+                                    if x is not None:
+                                        closest_gap_forward = x - self._lidar_timestamp_list[log][i]
+                                        assert closest_gap_forward > 0
+                                        closest_ts_forward = x
+                                        break
+                            if self.return_missing_image_as == "closest":
```
johnwlambert, line 220: i think a cleaner way to do this is to just call `get_closest_cam_channel_timestamp`, but with a flag that states `ignore_ts_threshold=True`. It would be a new default argument set to False usually
https://github.com/argoai/argoverse-api/blob/master/argoverse/data_loading/synchronization_database.py#L172

Then I think we can eliminate 15 lines of code here

johnwlambert: we would need to change `get_closest_cam_channel_timestamp()` so that if
```python
if ignore_ts_threshold == True:
    return closest_cam_ch_timestamp
```
immediately, instead of doing the timestamp diff checks

johnwlambert (changes requested):

PR #163, opened 2020-09-25
add intensity, ring index support in ply loader

johnwlambert:
Provides a function to load not only x,y,z coordinates for each point in an Argoverse LiDAR sweep, but also to load reflectance/intensity and the ring index.

Thanks to @irenecortes for this update from her repo (Ref: https://github.com/irenecortes/argoverse_lidar), and to @apak-00 for the function name.


benjaminrwilson:
Not sure what the status of this is -- but this looks promising as well: http://www.open3d.org/docs/latest/python_api/open3d.t.io.read_point_cloud.html#open3d.t.io.read_point_cloud (if this is ever considered being added as a dependency).

argoverse/utils/ply_loader.py
```diff
@@ -26,3 +26,23 @@ def load_ply(ply_fpath: _PathLike) -> np.ndarray:
     z = np.array(data.points.z)[:, np.newaxis]
 
     return np.concatenate((x, y, z), axis=1)
+
+
+def load_ply_xyzil(ply_fpath: _PathLike) -> np.ndarray:
```
johnwlambert, line 31: maybe pass in string "xyz" or "xyzi" to determine what channels you want

argoverse/utils/ply_loader.py
```diff
@@ -26,3 +26,23 @@ def load_ply(ply_fpath: _PathLike) -> np.ndarray:
...
+
+    Returns:
+        arr: Array of shape (N, 5), i.e. x,y,z,intensity,laser_number
+    """
+
+    data = pyntcloud.PyntCloud.from_file(os.fspath(ply_fpath))
+    x = np.array(data.points.x)[:, np.newaxis]
+    y = np.array(data.points.y)[:, np.newaxis]
+    z = np.array(data.points.z)[:, np.newaxis]
+    i = np.array(data.points.intensity)[:, np.newaxis]
```
johnwlambert, line 45: make list comprehension from dictionary

venice-erin:
hi, any reason this haven't been merged yet?

argoverse/utils/ply_loader.py
johnwlambert: added

argoverse/utils/ply_loader.py
johnwlambert: added

argoverse/utils/ply_loader.py
```diff
@@ -26,3 +26,37 @@ def load_ply(ply_fpath: _PathLike) -> np.ndarray:
     z = np.array(data.points.z)[:, np.newaxis]
 
     return np.concatenate((x, y, z), axis=1)
+
+
+def load_ply_by_attrib(ply_fpath: _PathLike, attrib_spec: str) -> Optional[np.ndarray]:
```
benjaminrwilson, line 31: Should we default `attrib_spec` to `attrib_spec = "xyzil"`?

johnwlambert: i like that

johnwlambert: updated

argoverse/utils/ply_loader.py
```diff
@@ -26,3 +26,37 @@ def load_ply(ply_fpath: _PathLike) -> np.ndarray:
...
+            x -> point x-coord
+            y -> point y-coord
+            z -> point z-coord
+            i -> point intensity/reflectance
+            l -> laser number of laser from which point was returned
+
+    Returns:
+        arr: Array of shape (N, C). If attrib_str is invalid, `None` will be returned
+    """
+    possible_attributes = ["x", "y", "z", "i", "l"]
```
benjaminrwilson, line 46: I wonder if we should do this with an `OrderedDict`? Probably doesn't matter too much, but we could do something like:

```
possible_attributes = OrderedDict({"x": None, "y": None, "z": None, "i": None, "l": None})
filtered_attributes = list(filter(lambda x: x in possible_attributes, list(attrib_spec)))
if not filtered_attributes:
    return None

data = pyntcloud.PyntCloud.from_file(os.fspath(ply_fpath))
return np.stack([getattr(data.points, name) for name in filtered_attributes], axis=1)
```

Just a thought, though.

benjaminrwilson: Actually, I think `OrderedDict` and just be a `Dict`.

benjaminrwilson (approved):
👍

johnwlambert:
> hi, any reason this haven't been merged yet?

Hi @venice-erin, we were still discussing the API a bit now but we'll merge it now after we've made some updates.

venice-erin:
Thanks :)

PR #166, opened 2020-10-14
plot object categories w/ cuboids

johnwlambert:
Renders object category in front of a green alpha-blended rectangle.


argoverse/data_loading/object_label_record.py
```diff
@@ -11,10 +11,22 @@
...
-from argoverse.utils.cv2_plotting_utils import draw_clipped_line_segment
+from argoverse.utils.cv2_plotting_utils import draw_clipped_line_segment, proj_cam_to_uv
 from argoverse.utils.se3 import SE3
 from argoverse.utils.transform import quat2rotmat
 
+BKGRND_RECT_OFFS_UP = 30 # px
+BKGRND_RECT_OFFS_DOWN = 10 # px
+
+BKGRND_RECT_OFFS_LEFT = 70 # px
+BKGRND_RECT_OFFS_RIGHT = 70 # px
```
benjaminrwilson, line 22: Maybe a comment here explaining these? I'd probably have to do a little digging to know what they are.

argoverse/data_loading/object_label_record.py
```diff
@@ -11,10 +11,22 @@
...
+
+WHITE_BGR = (255,255,255)
+EMERALD_RGB = (80, 220, 100)
+BKGRND_RECT_ALPHA = 0.45
+
+TEXT_OFFS_LEFT = 70 # px
+MAX_RANGE_THRESH_PLOT_CATEGORY = 50 # meters
```
benjaminrwilson, line 29: Should this be consistent with the range we consider for the challenges?

argoverse/data_loading/object_label_record.py
```diff
@@ -191,6 +203,25 @@ def draw_rect(selected_corners: np.array, color: Tuple[int, int, int]) -> None:
         # Draw rear (last 4 corners) in red
         draw_rect(corners[4:], colors[1][::-1])
 
+        # grab the top vertices
+        center_top = np.mean(corners[[0,1,4,5]], axis=0)
+        uv_ct, _, _, _ = proj_cam_to_uv(center_top.reshape(1, 3), camera_config)
+        uv_ct = uv_ct.squeeze()
+
+        if np.linalg.norm(center_top) < MAX_RANGE_THRESH_PLOT_CATEGORY and uv_ct[0] >= 0 and uv_ct[1] >= 0:
```
benjaminrwilson, line 211: Potentially make this a helper function which returns a boolean w/ a descriptive name?

argoverse/data_loading/object_label_record.py
```diff
@@ -207,6 +238,56 @@ def draw_rect(selected_corners: np.array, color: Tuple[int, int, int]) -> None:
         return img
 
 
+def add_text_cv2(img: np.ndarray, text: str, x: int, y: int, color: Tuple[int,int,int], thickness: int = 3) -> None:
+    """Add text to image using OpenCV. Color should be RGB order"""
```
benjaminrwilson, line 242: I think there's an anti-aliasing option in `cv`. Might make the text look a bit nicer -- just a thought.

argoverse/data_loading/object_label_record.py
```diff
@@ -207,6 +238,56 @@ def draw_rect(selected_corners: np.array, color: Tuple[int, int, int]) -> None:
...
+
+
+def draw_alpha_rectangle(
+    img: np.ndarray,
+    top_left: Tuple[int,int],
+    bottom_right: Tuple[int,int],
+    color_rgb: Tuple[int,int,int],
+    alpha: float
+):
+    """Alpha blend colored rectangle into image. Coords given as (x,y) tuples"""
```
benjaminrwilson, line 261: Should this have args in the docstring?

argoverse/data_loading/object_label_record.py
```diff
@@ -207,6 +238,56 @@ def draw_rect(selected_corners: np.array, color: Tuple[int, int, int]) -> None:
...
+    img_h, img_w, _ = img.shape
+    mask = np.zeros((img_h,img_w),dtype=np.uint8)
+    mask[top_left[1]:bottom_right[1],top_left[0]:bottom_right[0]] = 1
+    return vis_mask(img, mask, np.array(list(color_rgb[::-1])), alpha)
+
+
+def vis_mask(img: np.ndarray, mask: np.ndarray, col: Tuple[int,int,int], alpha: float=0.4):
```
benjaminrwilson, line 268: Was `black` run on this file? Some of the spacing looks a bit different than their styling.

johnwlambert: not yet, i'll run it now

argoverse/data_loading/object_label_record.py
johnwlambert: gets too messy beyond 50 m in my experience

argoverse/data_loading/object_label_record.py
johnwlambert: not sure, i think is self explanatory? let me know if not though

argoverse/data_loading/object_label_record.py
johnwlambert: that's a good idea, thanks

argoverse/data_loading/object_label_record.py
```diff
@@ -204,23 +213,15 @@ def draw_rect(selected_corners: np.array, color: Tuple[int, int, int]) -> None:
         draw_rect(corners[4:], colors[1][::-1])
 
         # grab the top vertices
-        center_top = np.mean(corners[[0,1,4,5]], axis=0)
+        center_top = np.mean(corners[[0, 1, 4, 5]], axis=0)
```
benjaminrwilson, line 216: I might consider a variable `TOP_VERT_INDICES: List[int] = [0, 1, 4, 5]`

argoverse/data_loading/object_label_record.py
```diff
@@ -204,23 +213,15 @@ def draw_rect(selected_corners: np.array, color: Tuple[int, int, int]) -> None:
...
         uv_ct, _, _, _ = proj_cam_to_uv(center_top.reshape(1, 3), camera_config)
-        uv_ct = uv_ct.squeeze()
+        uv_ct = uv_ct.squeeze().astype(np.int32)  # cast to integer
```
benjaminrwilson, line 218: This is probably outside the scope of this commit, but I wonder if we should have clear indication of whether we `floor` or `ceil` when converting to discrete coordinates (maybe as a flag to `proj_cam_to_uv`?

argoverse/visualization/vis_mask.py
```diff
@@ -37,17 +37,23 @@
 Segment = Tuple[float, float, float, float]
 
 
-def vis_mask(image: np.ndarray, mask: np.ndarray, color: float, alpha: float = 0.4) -> np.ndarray:
-    """Visualize a single binary mask.
+def vis_mask(
+    image: np.ndarray, mask: np.ndarray, color: Union[float, Tuple[int, int, int]], alpha: float = 0.4
```
benjaminrwilson, line 41: `color: Union[float, Tuple[int, int, int]]` -- does a single `float` represent greyscale?

argoverse/data_loading/object_label_record.py
johnwlambert: I think shouldn't matter much, since both floor or ceil can push float uv coords out of the image boundaries. i think closest (round to int), then value check is appropriate. but our `proj_cam_to_uv` is actually really messed up (not vectorized) which we need to fix in another PR

argoverse/data_loading/object_label_record.py
johnwlambert: good call, will do

argoverse/visualization/vis_mask.py
johnwlambert: that's right

argoverse/data_loading/object_label_record.py
johnwlambert: Added, thanks for the suggestion

argoverse/data_loading/object_label_record.py
johnwlambert: added a few comments, yeah that was not well-documented

benjaminrwilson:
LGTM 👍

PR #170, opened 2020-10-22
allow options on rendering centerlines, directly generating mp4, and …

johnwlambert:
For `cuboids_to_bboxes.py` script:
- allow options on rendering centerlines
- allow directly generation mp4 without individual frames
- use consistent notation for SE(3)
- add multiprocessing support


argoverse/utils/cv2_video_utils.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/python3
+
+import cv2
+import numpy as np
+
+"""
+Python-based utilities to avoid blowing up the disk with images, as FFMPEG requires.
```
benjaminrwilson, line 7: Does this replace this dependency? https://github.com/argoai/argoverse-api/blob/1dfc03ee8dc9b338eac6c86e21d7b999b59e2498/argoverse/visualization/generate_sequence_videos.py#L108

johnwlambert: I didn't realize we had this dependency tbh (it's not in our dependency list for some reason), I'll take a look

janesjanes (approved):
Looks good to me, the code run  without any problems

demo_usage/cuboids_to_bboxes.py
```diff
@@ -216,40 +238,64 @@ def dump_clipped_3d_cuboids_to_images(
...
+                args.max_num_images_to_render * 9,
+                args.dataset_dir,
+                args.experiment_prefix,
+                not args.no_motion_compensation,
+                args.omit_centerlines,
+                args.generate_video_only,
+            )
+            for log_id in log_ids
+        ]
+        with Pool(os.cpu_count()) as p:
```
janesjanes, line 286: Since it take awhile to run, would be nice to be able to specify number of cpu to use as well so we can do something else while rendering.

argoverse/utils/cv2_video_utils.py
benjaminrwilson: What were your thoughts on this?

johnwlambert: haven't played around with moviepy before, but we should check it out more: https://zulko.github.io/moviepy/

benjaminrwilson:
Do we have any blockers here?

benjaminrwilson:
LGTM 👍

PR #178, opened 2020-11-28
add more clear compose function that is alias for right_multiply_with…

johnwlambert:
Add a `compose()` method to the SE(3) class, which is an alias for `right_multiply_with_se3()` and has a more clear naming convention. In line with [GTSAM's `compose()`](https://github.com/borglab/gtsam/blob/develop/gtsam/geometry/Pose3.h#L111)

### DIFF
--- a/argoverse/utils/se3.py
+++ b/argoverse/utils/se3.py
@@ -61,10 +61,10 @@ class SE3:
         Algebraic representation: chained_se3 = T * right_se3
 
         Args:
-            right_se3: instance of SE3 class
+            right_se3: another instance of SE3 class
 
         Returns:
-            chained_se3: instance of SE3 class
+            chained_se3: new instance of SE3 class
         """
         chained_transform_matrix = self.transform_matrix.dot(right_se3.transform_matrix)
         chained_se3 = SE3(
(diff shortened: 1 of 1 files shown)


argoverse/utils/se3.py
```diff
@@ -55,20 +55,33 @@ def inverse(self) -> "SE3":
         """
         return SE3(rotation=self.rotation.T, translation=self.rotation.T.dot(-self.translation))
 
-    def right_multiply_with_se3(self, right_se3: "SE3") -> "SE3":
-        """Right multiply the transformation matrix with another SE3 instance.
+    def compose(self, right_se3: "SE3") -> "SE3":
+        """Compose (right multiply) this class' transformation matrix T with another SE3 instance.
```
benjaminrwilson, line 59: I might would say "by" instead of "with", but probably a matter of preference ...

argoverse/utils/se3.py
```diff
@@ -55,20 +55,33 @@ def inverse(self) -> "SE3":
...
 
         Args:
-            right_se3: instance of SE3 class
+            right_se3: another instance of SE3 class
 
         Returns:
-            chained_se3: instance of SE3 class
+            chained_se3: new instance of SE3 class
         """
         chained_transform_matrix = self.transform_matrix.dot(right_se3.transform_matrix)
```
benjaminrwilson, line 69: I know this wasn't changed with this PR, but any aversion to using the matrix multiplication operator introduced in Python 3.5 `self.transform_matrix @ right_se3.transform_matrix`? I think it's a bit more readable, especially when following algebraic notation.

johnwlambert: This is a nice suggestion, thanks. Done

janesjanes (approved):
look good!

PR #179, opened 2020-12-01
Update object_label_record.py

johnwlambert:


janesjanes:
looks good to me, namedtuple seems nice

PR #189, opened 2021-01-13
Added detection module init.

benjaminrwilson:
`pip install git+https://github.com/argoai/argoverse-api.git` does not include the `argoverse.evaluation.detection` module since it's not registered as a module. This should fix that and allow for remote installation.


johnwlambert:
Confirmed the fix

`import argoverse.evaluation.detection` works with 
`pip install git+https://github.com/benjaminrwilson/argoverse-api.git@detection-init`, but not with
`pip install git+https://github.com/argoai/argoverse-api.git`

PR #191, opened 2021-01-15
add script to convert tracking dataset to forecasting dataset

johnwlambert:
- Adds functionality to convert the 3d tracking dataset into a forecasting-style dataset (1 csv per log)

### DIFF
--- a/demo_usage/convert_tracking_dataset_to_forecasting.py
+++ b/demo_usage/convert_tracking_dataset_to_forecasting.py
@@ -1,0 +1,196 @@
+
+import copy
+import csv
+import pdb
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Tuple
+
+
+import numpy as np
+from argoverse.data_loading.simple_track_dataloader import SimpleArgoverseTrackingDataLoader
+from argoverse.data_loading.object_label_record import json_label_dict_to_obj_record
+from argoverse.utils.se3 import SE3
+from argoverse.utils.transform import quat2rotmat, quat_argo2scipy
+from recordclass import RecordClass
+from scipy.spatial.distance import cdist
+from scipy.spatial.transform import Rotation
+
+
+class Box3d(RecordClass):
+	center: np.ndarray # x, y, z
+	width: float
+	length: float
+	height: float
+	orientation_q: Tuple[float,float,float,float] # qw, qx, qy, qz
+	name: str
+	track_id: str
+	timestamp: int
+	orientation_yaw: Optional[float] = None
+	v0: Optional[Tuple[float,float,float]] = None # vertex 0
+	v1: Optional[Tuple[float,float,float]] = None # vertex 1
+	v2: Optional[Tuple[float,float,float]] = None # vertex 2
+	v3: Optional[Tuple[float,float,float]] = None # vertex 3
+
+
+def rotmat2quat(R: np.ndarray) -> np.ndarray:
+	"""  """
+	q_scipy =  Rotation.from_matrix(R).as_quat()
+	return quat_scipy2argo(q_scipy)
+
+
+def quat_scipy2argo(q_scipy: np.ndarray) -> np.ndarray:
+	"""Re-order Argoverse's scalar-first [w,x,y,z] quaternion order to Scipy's scalar-last [x,y,z,w]"""
+	x, y, z, w = q_scipy
+	q_argo = np.array([w, x, y, z])
+	return q_argo
+
+
+def quaternion_yaw_scipy(q: Tuple[float,float,float,float]) -> float:
+    """ """
+    q_argo = q
+    q_scipy = quat_argo2scipy(q_argo)
+    yaw, _, _ = Rotation.from_quat(q_scipy).as_euler('zyx')
+    return yaw
+
+
+
+
+def remove_duplicate_instances(instances: List[Box3d]) -> List[Box3d]:
+	"""Remove any duplicate cuboids in ground truth.
+	Any ground truth cuboid of the same object class that shares the same centroid
+	with another is considered a duplicate instance.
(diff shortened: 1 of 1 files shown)


demo_usage/convert_tracking_dataset_to_forecasting.py
```diff
@@ -0,0 +1,196 @@
...
+	timestamp: int
+	orientation_yaw: Optional[float] = None
+	v0: Optional[Tuple[float,float,float]] = None # vertex 0
+	v1: Optional[Tuple[float,float,float]] = None # vertex 1
+	v2: Optional[Tuple[float,float,float]] = None # vertex 2
+	v3: Optional[Tuple[float,float,float]] = None # vertex 3
+
+
+def rotmat2quat(R: np.ndarray) -> np.ndarray:
+	"""  """
```
wqi, line 36: [Nit] Should we add a brief docstring for these utility functions?

demo_usage/convert_tracking_dataset_to_forecasting.py
```diff
@@ -0,0 +1,196 @@
...
+from argoverse.data_loading.object_label_record import json_label_dict_to_obj_record
+from argoverse.utils.se3 import SE3
+from argoverse.utils.transform import quat2rotmat, quat_argo2scipy
+from recordclass import RecordClass
+from scipy.spatial.distance import cdist
+from scipy.spatial.transform import Rotation
+
+
+class Box3d(RecordClass):
+	center: np.ndarray # x, y, z
```
wqi, line 20: If `center`s are always `np.ndarrays` of 3 elements, should we consider enforcing this with a type of `Tuple[float, float, float]`?

demo_usage/convert_tracking_dataset_to_forecasting.py
```diff
@@ -0,0 +1,196 @@
...
+	# choose the first instance in each column where repeat occurs
+	first_col_idxs = col_idxs[unique_element_idxs]
+	# eliminate redundant column indices
+	unique_ids = np.unique(first_col_idxs)
+	return instances[unique_ids]
+
+
+def compute_affinity_matrix(dts1: List[Box3d], dts2: List[Box3d]) -> np.ndarray:
+	"""Calculate the affinity matrix between detections and ground truth labels,
+	using a specified affinity function type.
```
wqi, line 91: [Nit] Does this function take a configurable affinity function type? If not, we should consider updating this docstring.

demo_usage/convert_tracking_dataset_to_forecasting.py
```diff
@@ -0,0 +1,196 @@
...
+    """ """
+    q_argo = q
+    q_scipy = quat_argo2scipy(q_argo)
+    yaw, _, _ = Rotation.from_quat(q_scipy).as_euler('zyx')
+    return yaw
+
+
+
+
+def remove_duplicate_instances(instances: List[Box3d]) -> List[Box3d]:
```
wqi, line 58: [Nit] Could use a more generic input type for the args if there's no specific need for a `List`. A `Sequence` should be ok here.

demo_usage/convert_tracking_dataset_to_forecasting.py
```diff
@@ -0,0 +1,196 @@
...
+			orientation_q = [qw, qx, qy, qz],
+			name = label['label_class'],
+			track_id = label['track_label_uuid'],
+			timestamp = timestamp
+		)
+
+		# transform box from the egovehicle frame into the city frame
+		box.center = city_SE3_egovehicle.transform_point_cloud(np.array(box.center).reshape(1,3)).squeeze()
+		box.center = list(box.center)
+		box.orientation_q = list(rotmat2quat(city_SE3_egovehicle.rotation @ quat2rotmat(list(box.orientation_q))))
```
wqi, line 142: Maybe I'm missing something here, but doesn't L24 specify `orientation_q` as a Tuple?

demo_usage/convert_tracking_dataset_to_forecasting.py
```diff
@@ -0,0 +1,196 @@
...
+		box.v1 = list(vertices_cityfr[1])
+		box.v2 = list(vertices_cityfr[2])
+		box.v3 = list(vertices_cityfr[3])
+		box.orientation_yaw = quaternion_yaw_scipy(box.orientation_q)
+		box_list.append(box)
+
+	return box_list
+
+
+def create_tracks_csv(root_path: str, split_subdir: str) -> None:
```
wqi, line 154: [Nit] Generally, we should prefer to use pathlib `Path` objects in place of strings. If the rest of the argoverse-api uses `str`s, then let's just leave this as is.

demo_usage/convert_tracking_dataset_to_forecasting.py
```diff
@@ -0,0 +1,196 @@
...
+		fieldnames = dict_list[0].keys()
+		writer = csv.DictWriter(csvfile, fieldnames=fieldnames, delimiter=delimiter)
+		writer.writeheader()
+		for row_dict in dict_list:
+			writer.writerow(row_dict)
+
+
+if __name__ == '__main__':
+
+	root_path = '/home/ubuntu/argoverse/argoverse-tracking/val-1.1/argoverse-tracking'
```
wqi, line 193: [Nit] I'm guessing the root path will need to be changed by the end user if we don't provide a CLI arg to do so. Maybe we should call out all the places the end user might need to change variables with all caps comments.

demo_usage/convert_tracking_dataset_to_forecasting.py
```diff
@@ -0,0 +1,196 @@
+
```
wqi, line 1: [Nit] Empty line

wqi (approved):
Thanks for adding this! Functionality looks good, I just left a couple nits for maintainability.

PR #195, opened 2021-02-16
Require manual specification of number of processes to use in detection eval

johnwlambert:
Prevent issues in docker container where os.cpu_count() can have unintended consequences


benjaminrwilson:
LGTM 👍

PR #198, opened 2021-03-01
Adding map_files_root as an argument of ArgoverseMap

venice-liong:
In this PR we add `map_files_root` in `ArgoverseMap` during initialization to make it flexible in accessing maps at different paths.

The default remains to be `$ROOT/map_files` to ensure backward compatibility.

### DIFF
--- a/argoverse/map_representation/map_api.py
+++ b/argoverse/map_representation/map_api.py
@@ -54,8 +54,14 @@ class ArgoverseMap:
     are not provided, but can be hallucinated if one considers an average lane width.
     """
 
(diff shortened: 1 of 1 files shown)


johnwlambert:
Hi @venice-liong, thanks for your PR and sorry for the slow review.

We ended up needing to make a few small changes to this PR for the CI, and to support `Path` instead of just `str` arguments, so we put them in this PR https://github.com/argoai/argoverse-api/pull/207 which was just merged. I'll go ahead and close this one, for now.

Thank you again and please let us know if you run into any issues.

PR #201, opened 2021-04-01
add Sim(2)

johnwlambert:
Add support for 2d rigid body transformations with scaling, e.g. Similarity(2), instead of just SE(2) transformations. SE(2) functionality is already implemented in `argoverse-api` [here](https://github.com/argoai/argoverse-api/blob/master/argoverse/utils/se2.py)


argoverse/utils/sim2.py
```diff
@@ -0,0 +1,95 @@
...
+"""
+
+from typing import Union
+
+import numpy as np
+
+from argoverse.utils.helpers import assert_np_array_shape
+
+
+class Sim2:
```
benjaminrwilson, line 16: Should we potentially create a `LieGroup` super-class? Seems like there would be a lot of overlapping functionality between these different transformations.

benjaminrwilson: Additionally, I think sub-classing `numpy.ndarray` might be something we should evaluate: https://numpy.org/doc/stable/user/basics.subclassing.html. We could implement matrix multiplication directly on the object itself (in addition to transformation methods).

argoverse/utils/sim2.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if not np.allclose(self.rotation(), other.rotation()):
+            return False
+
+        if not np.allclose(self.translation(), other.translation()):
+            return False
+
+        return True
+
+    def rotation(self) -> np.ndarray:
```
benjaminrwilson, line 47: Might consider making a `@property`?

argoverse/utils/sim2.py
```diff
@@ -0,0 +1,95 @@
...
+        """Composition with another Sim2."""
+        return Sim2(self.R_ * S.R_, ((1.0 / S.s_) * self.t_) + self.R_ @ S.t_, self.s_ * S.s_)
+
+    def inverse(self) -> "Sim2":
+        """Return the inverse."""
+        Rt = self.R_.T
+        sRt = -Rt @ (self.s_ * self.t_)
+        return Sim2(Rt, sRt, 1.0 / self.s_)
+
+    def transformFrom(self, point_cloud: np.ndarray) -> np.ndarray:
```
benjaminrwilson, line 77: I imagine this naming is by convention, but I personally get confused by it. What would be positive / negatives of something like `to_target` or `to_target_frame` vs `transformFrom`?

Additionally, depending on the discussion above, should we consider conforming to snake case?

johnwlambert: Just switched to snake case -- thanks.

argoverse/utils/sim2.py
johnwlambert: thanks, just added it

argoverse/utils/sim2.py
johnwlambert: Haven't seen the subclassing before on `numpy.ndarray` -- would it make things more efficient? i'm not sure it would be more readable since there are a few different types of matrices here (3x3, or 3x1, or 4x4)

argoverse/utils/sim2.py
```diff
@@ -0,0 +1,106 @@
+"""
+Utility for 2d rigid body transformations with scaling.
+
+Refs:
+    http://ethaneade.com/lie_groups.pdf
+    https://github.com/borglab/gtsam/blob/develop/gtsam_unstable/geometry/Similarity3.h
```
benjaminrwilson, line 6: This link gives me a 404.

argoverse/utils/sim2.py
benjaminrwilson: I don't think we'd see any performance improvement. It would mainly be syntactic sugar, but I guess it could get confusing and lead to unintentional behavior. Idk, I just think that transformation syntax gets pretty verbose.

benjaminrwilson: Another thought -- LieGroups seems like they should almost be an entirely separate library.

argoverse/utils/sim2.py
```diff
@@ -0,0 +1,106 @@
...
+
+        Args:
+            R: array of shape (2x2) representing 2d rotation matrix
+            t: array of shape (2,) representing 2d translation
+            s: scaling factor
+        """
+        assert_np_array_shape(R, (2, 2))
+        assert_np_array_shape(t, (2,))
+        assert isinstance(s, float) or isinstance(s, int)
+        self.R_ = R.astype(np.float32)
```
benjaminrwilson, line 30: Should we make this casting behavior more visible to the user? What if they intended to use `float64`?

argoverse/utils/sim2.py
```diff
@@ -0,0 +1,106 @@
...
+        self.t_ = t.astype(np.float32)
+        self.s_ = float(s)
```
benjaminrwilson, line 32: Does `float` cast to a 64-bit value?

argoverse/utils/sim2.py
```diff
@@ -0,0 +1,106 @@
...
+        """Return the scale."""
+        return self.s_
+
+    @property
+    def matrix(self) -> np.ndarray:
+        """Calculate 3*3 matrix group equivalent"""
+        T = np.zeros((3, 3))
+        T[:2, :2] = self.R_
+        T[:2, 2] = self.t_
+        T[2, 2] = 1 / self.s_
```
benjaminrwilson, line 71: We might want to validate that `self.s_` does not equal zero in the constructor to avoid `ZeroDivisionError`.

argoverse/utils/sim2.py
```diff
@@ -0,0 +1,106 @@
...
+        """Composition with another Sim2."""
+        return Sim2(self.R_ * S.R_, ((1.0 / S.s_) * self.t_) + self.R_ @ S.t_, self.s_ * S.s_)
+
+    def inverse(self) -> "Sim2":
+        """Return the inverse."""
+        Rt = self.R_.T
+        sRt = -Rt @ (self.s_ * self.t_)
+        return Sim2(Rt, sRt, 1.0 / self.s_)
+
+    def transform_from(self, point_cloud: np.ndarray) -> np.ndarray:
```
benjaminrwilson, line 84: Should `point_cloud` be a more generic argument name?

tests/test_sim2.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    for sample in range(1000):
+        # generate random 2x2 rotation matrix
+        theta = np.random.rand() * 2 * np.pi
+        R = rotmat2d(theta)
+        t = np.random.randn(2)
+
+        pts_b = np.random.randn(25, 2)
+
+        aTb = SE2(copy.deepcopy(R), copy.deepcopy(t))
```
benjaminrwilson, line 178: I've noticed a lot of `deepcopy` ops throughout the repository. Do you think this is something that should be baked into the class itself? Would the end user ever want to construct an object without reallocating one of its args?

benjaminrwilson (commented):
Looks good, John. I had a few thoughts regarding overall structure in the code base moving forward. Definitely curious on your thoughts.

argoverse/utils/sim2.py
johnwlambert: good catch -- i just updated it with the correct link

argoverse/utils/sim2.py
johnwlambert: very good point -- i just updated the constructor to throw an exception in that case. and i added a new unit test to check that case also

argoverse/utils/sim2.py
johnwlambert: it could be, e.g. `points`. But currently I've just kept it consistent with the function signature for SE(2) and SE(3)

tests/test_sim2.py
johnwlambert: this is just the unit test -- to make sure that both functions see the exact same input (when asserting they create the same output).

If it was part of the class instead, yeah would make sense to bake it in somehow

argoverse/utils/sim2.py
johnwlambert: Hmm I was using this line to account just for the case when the scale `s` is an integer, and we want to cast it to a float

benjaminrwilson:
LGTM 👍

PR #206, opened 2021-04-12
add supporting code for the stereo competition

jhonykaesemodel:
Add supporting code for the stereo competition. New stereo data loader, stereo evaluation, Jupyter notebook for tutorial, etc.


argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+
+import cv2
+import disparity_interpolation
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+
+logger = logging.getLogger(__name__)
+
+log_colormap = [
```
johnwlambert, line 62: Thanks for the PR @jhonykaesemodel , looks nice. I'll leave a few comments as I work my way through it.

do you mind adding a comment explaining what the `log_colormap` is used for? and maybe a quick equation for it inline?

argoverse/evaluation/stereo/disparity_interpolation/disparity_interpolation/source/disparity_interpolation.pyx
```diff
@@ -0,0 +1,40 @@
+# distutils: language = c++
+# distutils: sources = nn_interpolation.cpp
+# cython: language_level=3
+
+import cython
+
+import numpy as np
+cimport numpy as np
+
+
```
johnwlambert, line 10: How does this interpolation speed compare to what sklearn or scipy provides? Maybe you could add in the comment what the expected run time is for a typical stereo image, e.g. 2056 × 2464?

Would this be faster if we wrapped it in Pybind in pure C++ instead of Cython?

Would it possibly make sense to release this as a separate repo that we could put on pybind, so that nothing has to be compiled?

argoverse/evaluation/stereo/disparity_interpolation/disparity_interpolation/source/nn_interpolation.cpp
```diff
@@ -0,0 +1,80 @@
+#include "nn_interpolation.h"
+#include <stdint.h>
+
```
johnwlambert, line 3: maybe a 3-4 line docstring at the top of this file could describe the high-level algorithm/pseudocode?

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,432 @@
...
+            name: disparity map name, one of:
+
+                ["stereo_front_left_rect_disparity",
+                 "stereo_front_left_rect_objects_disparity"]
+
+            log_id: log_id, if not specified will use self.current_log
+            load: whether to return image array (True) or image path (False)
+
+        Returns:
+            np.array: list of disparity image path (str or np.array)),
```
johnwlambert, line 295: i think this return type `np.array` classes with the `List[str]` return type -- which is preferred?

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+
+def compute_disparity_error(
+    pred_fpath: Path,
+    gt_fpath: Path,
+    gt_obj_fpath: Path,
+    abs_error_thresholds: List = [10, 5, 3],
+    rel_error_thresholds: List = [0.1, 0.1, 0.1],
+    figs_fpath: Path = None,
+    save_disparity_error_image: bool = False,
+):
```
johnwlambert, line 188: nit: missing return type hint

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+            summary[f"bg*:{str(abs_error_threshold)}"] = d1_bg_est * 100
+
+        return summary
+
+
+def compute_disparity_error(
+    pred_fpath: Path,
+    gt_fpath: Path,
+    gt_obj_fpath: Path,
+    abs_error_thresholds: List = [10, 5, 3],
```
johnwlambert, line 184: maybe `[10, 5, 3],` should be named defaults at the top of the file?
```python
# absolute error thresholds in pixels, also used for KITTI stereo eval
DEFAULT_ABS_ERROR_THRESHOLDS = [10,5,3]
```

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+    rel_error_thresholds: List = [0.1, 0.1, 0.1],
```
johnwlambert, line 185: same not here -- maybe a comment at the top of the file could explain why `rel_error_thresholds` are useful? to make it self-contained for the reader?

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+    for abs_error_threshold, rel_error_threshold in zip(abs_error_thresholds, rel_error_thresholds):
+        bad_pixels = (abs_err > abs_error_threshold) & (rel_err > rel_error_threshold)
+
+        errors[f"num_errors_bg:{str(abs_error_threshold)}"] = np.sum(bg_mask & bad_pixels)
+        errors[f"num_errors_fg:{str(abs_error_threshold)}"] = np.sum(fg_mask & bad_pixels)
+        errors[f"num_errors_bg_est:{str(abs_error_threshold)}"] = np.sum(bg_mask & pred_mask & bad_pixels)
+        errors[f"num_errors_fg_est:{str(abs_error_threshold)}"] = np.sum(fg_mask & pred_mask & bad_pixels)
+
+    if save_disparity_error_image:
+        err = np.minimum(abs_err / abs_error_thresholds[0], rel_err / rel_error_thresholds[0])
```
johnwlambert, line 233: this function is getting quite long; should lines 233 - 281 go inside a helper function?

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+        log_id = str(gt_fpath).split("/")[-3]
+        timestamp = str(gt_fpath).split("_")[-1][:-4]
+        save_dir = f"{figs_fpath}/{log_id}/"
+        Path(save_dir).mkdir(parents=True, exist_ok=True)
+        cv2.imwrite(f"{save_dir}/disparity_error_{timestamp}.png", d_img)
+
+    return errors
+
+
+def accumulator(abs_error_thresholds: List) -> pd.DataFrame:
```
johnwlambert, line 286: I might prefer to name this `accumulate_stereo_metrics()` or something since `accumulator` seems like a class name.

argoverse/utils/camera_stats.py
```diff
@@ -25,7 +25,12 @@
     "ring_side_right",
 ]
 
-STEREO_CAMERA_LIST = ["stereo_front_left", "stereo_front_right"]
+STEREO_CAMERA_LIST = [
+    "stereo_front_left",
+    "stereo_front_right",
+    "stereo_front_left_rect",
```
johnwlambert, line 31: is this new rectified imagery we are releasing, in new zip files?

We do use this variable in some places, and changing like this without a new dataset release would break some of our other demo scripts, so I might prefer to make a new variable

`RECTIFIED_STEREO_CAMERA_LIST = [ blah, blah ] `

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+        sizes = np.logspace(8, 12, num=10, base=2.0)
+        scalars = np.array(log_colormap, dtype=object)[:, 0] * abs_error_thresholds[0]
+        scalars = [f"[{scalar[0]:.2f}, {scalar[1]:.2f}]" for scalar in scalars]
+        colors = np.array(log_colormap, dtype=object)[:, 1]
+        colors = [color / 255.0 for color in colors]
+        plt.scatter(x, y, c=colors, s=3000)
+        plt.xlabel("Disparity range errors in pixels", fontsize=24)
+        plt.xticks(x, scalars, fontsize=20)
+        plt.yticks([])
+        ax.spines["top"].set_visible(False)
```
johnwlambert, line 259: inline comments might make this a bit clearer -- I haven't used the `spines` attribute before

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+                data_sum[f"num_errors_bg:{str(abs_error_threshold)}"]
+                + data_sum[f"num_errors_fg:{str(abs_error_threshold)}"]
+            ) / (data_sum["num_pixels_bg"] + data_sum["num_pixels_fg"])
+            d1_fg = data_sum[f"num_errors_fg:{str(abs_error_threshold)}"] / data_sum["num_pixels_fg"]
+            d1_bg = data_sum[f"num_errors_bg:{str(abs_error_threshold)}"] / data_sum["num_pixels_bg"]
+            d1_all_est = (
+                data_sum[f"num_errors_bg_est:{str(abs_error_threshold)}"]
+                + data_sum[f"num_errors_fg_est:{str(abs_error_threshold)}"]
+            ) / (data_sum["num_pixels_bg_est"] + data_sum["num_pixels_fg_est"])
+            d1_fg_est = data_sum[f"num_errors_fg_est:{str(abs_error_threshold)}"] / data_sum["num_pixels_fg_est"]
```
johnwlambert, line 167: is the `str` necessary in this f-string?
 `:{str(abs_error_threshold)}` 

wont that cast occur by default?

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+                for (pred_fpath, gt_fpath, gt_obj_fpath) in zip(pred_fpaths, gt_fpaths, gt_obj_fpaths)
+            ]
+            with Pool(self.num_procs) as p:
+                errors = p.starmap(compute_disparity_error, args)
+
+        data = pd.concat(errors)
+        data_sum = data.sum()
+        summary: dict() = dict()
+
+        for abs_error_threshold in self.abs_error_thresholds:
```
johnwlambert, line 156: maybe `abs_error_thresh` is more readable than `abs_error_threshold`; getting a bit long-winded for variable names

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+        self.gt_root_fpath = gt_root_fpath
+        self.figs_fpath = figs_fpath
+        self.abs_error_thresholds = abs_error_thresholds
+        self.rel_error_thresholds = rel_error_thresholds
+        self.save_disparity_error_image = save_disparity_error_image
+        self.num_procs = os.cpu_count() if num_procs == -1 else num_procs
+
+    def evaluate(self) -> pd.DataFrame:
+        """Evaluate stereo output and return metrics. The multiprocessing
+        library is used for parallel processing of disparities.
```
johnwlambert, line 107: Maybe we could clarify what `parallel processing of disparities` means -- `parallel processing of disparity evaluation`?

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
+# <Copyright 2021, Argo AI, LLC. Released under the MIT license.>
+"""Stereo evaluation for the Argoverse stereo leaderboard.
+
+Evaluation:
+
+    We consider the disparity of a pixel to be correctly estimated if the absolute disparity error is less than a threshold and its relative error is less than 10% of its true value, similar to the KITTI Stereo 2015 benchmark [1]. We define three disparity error thresholds: 3, 5, and 10 pixels.
+
+    Some stereo matching methods such as Semi-Global Matching (SGM) might provide sparse disparity maps, meaning that some pixels will not have valid disparity values. In those cases, we interpolate the predicted disparity map using a simple nearest neighbor interpolation as in the KITTI Stereo 2015 benchmark [1] to assure we can compare it to our semi-dense ground-truth disparity map. Current deep stereo matching methods normally predict disparity maps with 100% density. Thus, an interpolation step is not needed for the evaluation.
```
johnwlambert, line 8: line length is a bit long -- our CI will enforce <120 chars per line

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,432 @@
...
+    def get_image_list(self, camera: str, log_id: Optional[str] = None, load: bool = False) -> List[str]:
+        """Get list of image/or image path
+
+        Args:
+            camera: camera based on camera_stats.STEREO_CAMERA_LIST
+            log_id: log_id, if not specified will use self.current_log
+            load: whether to return image array (True) or image path (False)
+
+        Returns:
+            np.array: list of image path (str or np.array)),
```
johnwlambert, line 270: same note here about return type

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,432 @@
...
+        assert self.image_list is not None
+        assert self._image_list is not None
+
+        if self._image_timestamp_list is None:
+            self._image_timestamp_list = {}
+            for log in self.log_list:
+                self._image_timestamp_list[log] = {}
+                for camera in STEREO_CAMERA_LIST:
+                    self._image_timestamp_list[log][camera] = [
+                        int(x.split("/")[-1][:-4].split("_")[-1]) for x in self._image_list[log][camera]
```
johnwlambert, line 177: maybe cleaner as `int(Path(x).stem.split("_")[-1])` than `int(x.split("/")[-1][:-4].split("_")[-1])`

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,432 @@
...
+                                )
+                            )
+                        )
+                    )
+                    self.image_count += len(self._image_list[log][camera])
+        return self._image_list[self.current_log]
+
+    @property
+    def disparity_list(self) -> Dict[str, List[str]]:
+        """return list of all image path (str) for all cameras for the current log
```
johnwlambert, line 138: typo "image paths"

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,432 @@
...
+                x
+                for x in os.listdir(os.path.join(self.root_dir, "rectified_stereo_images_v1.1", self.split_name))
+                if valid_log(x)
+            ]
+
+        return self._log_list
+
+    @property
+    def image_list(self) -> Dict[str, List[str]]:
+        """return list of all image path (str) for the current log
```
johnwlambert, line 114: is the doc string correct here -- we loop over `self.log_list` but the docstring says its just for the current log

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,432 @@
...
+                self._calib[log] = load_calib(calib_filename)
+
+    @property
+    def log_list(self) -> List[str]:
+        """return list of log (str) in the current dataset directory
+
+        Returns:
+            log_list: list of string representing log id
+        """
+        if self._log_list is None:
```
johnwlambert, line 91: prefer to return early rather than nested logic -- 
```python
if self._log_list is not None:
     return 
def valid_log(log: str) -> bool:
        return os.path.exists(
                    os.path.join(
                        self.root_dir,
                        "rectified_stereo_images_v1.1",
                        self.split_name,
                        log,
                        "vehicle_calibration_stereo_info.json",
                    )
                )

            self._log_list = [
                x
                for x in os.listdir(os.path.join(self.root_dir, "rectified_stereo_images_v1.1", self.split_name))
                if valid_log(x)
            ]
```

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,432 @@
...
+        """get calibration dict for current log
+
+        Returns:
+            calib: Calibration object for the current log
+        """
+        self._ensure_calib_is_populated()
+        assert self._calib is not None
+        return self._calib[self.current_log]
+
+    def _ensure_calib_is_populated(self) -> None:
```
johnwlambert, line 66: i believe double underscore is preferred for python private method

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,432 @@
...
+class ArgoverseStereoLoader:
+    def __init__(self, root_dir: str, split_name: str) -> None:
+        # initialize class member
+        self.STEREO_CAMERA_LIST = STEREO_CAMERA_LIST
+        self._log_list: Optional[List[str]] = None
+        self._image_list: Optional[Dict[str, Dict[str, List[str]]]] = None
+        self._disparity_list: Optional[Dict[str, Dict[str, List[str]]]] = None
+        self._image_timestamp_list: Optional[Dict[str, Dict[str, List[int]]]] = None
+        self._timestamp_image_dict: Optional[Dict[str, Dict[str, Dict[int, str]]]] = None
+        self._calib: Optional[Dict[str, Dict[str, Calibration]]] = None  # { log_name: { camera_name: Calibration } }
```
johnwlambert, line 29: is `# { log_name: { camera_name: Calibration } }` this commented out code intentionally left in?

johnwlambert:
I believe some tests are failing in CI:
```

=================================== FAILURES ===================================
__________________________ test_image_timestamp_list ___________________________

data_loader = <argoverse.data_loading.argoverse_stereo_loader.ArgoverseStereoLoader object at 0x7fb572254780>

    def test_image_timestamp_list(data_loader: ArgoverseStereoLoader) -> None:
        assert set(data_loader.image_timestamp_list.keys()) == set(STEREO_CAMERA_LIST)
        for camera in STEREO_CAMERA_LIST[3:]:
>           assert 1 in data_loader.image_timestamp_list[camera]
E           assert 1 in []

tests/test_argoverse_stereo_loader.py:38: AssertionError
__________________________ test_timestamp_image_dict ___________________________

data_loader = <argoverse.data_loading.argoverse_stereo_loader.ArgoverseStereoLoader object at 0x7fb572278748>

    def test_timestamp_image_dict(data_loader: ArgoverseStereoLoader) -> None:
        assert set(data_loader.timestamp_image_dict.keys()) == set(STEREO_CAMERA_LIST)
        for camera in STEREO_CAMERA_LIST[3:]:
>           assert len(data_loader.timestamp_image_dict[camera]) == 1
E           assert 0 == 1
E            +  where 0 = len({})

tests/test_argoverse_stereo_loader.py:44: AssertionError
_________________________ test_loading_image_disparity _________________________

data_loader = <argoverse.data_loading.argoverse_stereo_loader.ArgoverseStereoLoader object at 0x7fb571d06f98>

    def test_loading_image_disparity(data_loader: ArgoverseStereoLoader) -> None:
        camera = STEREO_CAMERA_LIST[2]
        disparity = STEREO_DISPARITY_LIST[0]
        disparity_objects = STEREO_DISPARITY_LIST[1]
        log = "1"
        image_1_at_timestamp = data_loader.get_image_at_timestamp(1, camera, log)
>       image_1 = data_loader.get_image(0, camera, log)

tests/test_argoverse_stereo_loader.py:65: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <argoverse.data_loading.argoverse_stereo_loader.ArgoverseStereoLoader object at 0x7fb571d06f98>
idx = 0, camera = 'stereo_front_left_rect', log_id = '1', load = True

    def get_image(
        self, idx: int, camera: str, log_id: Optional[str] = None, load: bool = True
    ) -> Union[str, np.ndarray]:
        """get image or image path at a specific index (in image index)
    
        Args:
            idx: image based 0-index
            camera: camera based on camera_stats.STEREO_CAMERA_LIST
            log_id: log_id, if not specified will use self.current_log
            load: whether to return image array (True) or image path (False)
    
        Returns:
            np.array: list of image path (str or np.array)),
        """
        assert self.image_timestamp_list is not None
        assert self._image_timestamp_list is not None
        assert self.image_list is not None
        assert self._image_list is not None
    
        if log_id is None:
            log_id = self.current_log
    
>       assert idx < len(self._image_timestamp_list[log_id][camera])
E       AssertionError

argoverse/data_loading/argoverse_stereo_loader.py:363: AssertionError
```

johnwlambert:
@benjaminrwilson maybe you can take a look too?

jhonykaesemodel:
Thanks a lot so far, @johnwlambert. Awesome comments! I'll address them asap, and sorry for this huge PR.

benjaminrwilson (commented):
Could we add an `__init__.py` file to `argoverse/data_loading/stereo` so it's found as a submodule?

benjaminrwilson:
Additionally, what you do you think about making a top-level `csrc` directory for c++ and cuda extensions? (e.g., https://github.com/rusty1s/pytorch_cluster/tree/master/csrc)

argoverse/evaluation/stereo/eval.py
benjaminrwilson: Additionally, what do you think about renaming `log_colormap` to `LOG_COLORMAP`. Also, maybe we should consider various colormaps in a task agnostic file (e.g., in some utilities section of `argoverse`)?

jhonykaesemodel:
```python
# The disparity error image uses a custom log-color scale depicting correct estimates in blue and wrong estimates in
# red color tones, as in the KITTI Stereo 2015 Benchmark [1].
# LOG_COLORMAP = [disparity error range, RGB color], where the disparity error range is defined as:
# y = 2^x for x = [-Inf, -4, -3, -2, -1, 0, 1, 2, 3, 4, Inf].
LOG_COLORMAP = [
    [np.array([0, 2 ** -4]), np.array([49, 54, 149], dtype=np.uint8)],
    [np.array([2 ** -4, 2 ** -3]), np.array([69, 117, 180], dtype=np.uint8)],
    [np.array([2 ** -3, 2 ** -2]), np.array([116, 173, 209], dtype=np.uint8)],
    [np.array([2 ** -2, 2 ** -1]), np.array([171, 217, 233], dtype=np.uint8)],
    [np.array([2 ** -1, 2 ** 0]), np.array([224, 243, 248], dtype=np.uint8)],
    [np.array([2 ** 0, 2 ** 1]), np.array([254, 224, 144], dtype=np.uint8)],
    [np.array([2 ** 1, 2 ** 2]), np.array([253, 174, 97], dtype=np.uint8)],
    [np.array([2 ** 2, 2 ** 3]), np.array([244, 109, 67], dtype=np.uint8)],
    [np.array([2 ** 3, 2 ** 4]), np.array([215, 48, 39], dtype=np.uint8)],
    [np.array([2 ** 4, np.Inf]), np.array([165, 0, 38], dtype=np.uint8)],
]
```

argoverse/evaluation/stereo/disparity_interpolation/disparity_interpolation/source/disparity_interpolation.pyx
jhonykaesemodel: Using the Scipy's interpolate.NearestNDInterpolator function it takes about 5 seconds. Using this Cython function it takes about 50 ms. It's 100x faster. 

Maybe pybind11 would be slightly faster. But I haven't yet tested.
I liked the idea of putting it on pypi, so we don't have to compile anything. So I removed the disparity_interpolation module and placed it on pypi. Now we simply need to `pip install disparity_interpolation`.

argoverse/evaluation/stereo/disparity_interpolation/disparity_interpolation/source/nn_interpolation.cpp
jhonykaesemodel: It's on pypi now.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Sure, changed to: 
```python
        Returns:
            List[str]: list of disparity image paths (str),
```

argoverse/evaluation/stereo/eval.py
jhonykaesemodel:
```python
    save_disparity_error_image: bool = False,
) -> pd.DataFrame:
```

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Agreed. Already changed.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Overlooked this. Removed all ```str``` from f-strings.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Yeah, changed to your suggestion.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Changed to "...parallel processing of disparity evaluation".

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Definitely. Changed.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Done.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Yes. Change to double score. But I think single underscore is used throughout the Argoverse API. We'd need to check on that later to be consistent.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Removed it.

argoverse/evaluation/stereo/disparity_interpolation/disparity_interpolation/source/nn_interpolation.cpp
johnwlambert: Wow that was fast! Really nice work.

I couldnt access https://github.com/jhonykaesemodel/disparity_interpolation though -- is that not pushed quite yet?

argoverse/utils/camera_stats.py
jhonykaesemodel: Agreed. I've changed this and also I created a new function ```load_stereo_calib``` in the ```calibration.py``` so I don't need to mess up with any other demos and tests.

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,367 @@
...
+            f"num_errors_bg:{abs_error_thresh}",
+            f"num_errors_fg:{abs_error_thresh}",
+            f"num_errors_bg_est:{abs_error_thresh}",
+            f"num_errors_fg_est:{abs_error_thresh}",
+        ]
+
+    return pd.DataFrame([[0] * num_fields], columns=columns)
+
+
+def interpolate_disparity(disp: np.array) -> np.array:
```
johnwlambert, line 325: Looks good. Type here should be `np.ndarray`

johnwlambert: do we have a unit test on this function? or in your other repo?

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,367 @@
...
+
+# relative error thresholds in pixels, also used for KITTI stereo eval
+DEFAULT_REL_ERROR_THRESHOLDS = [0.1, 0.1, 0.1]
+
+
+# The disparity error image uses a custom log-color scale depicting correct estimates in blue and wrong estimates in
+# red color tones, as in the KITTI Stereo 2015 Benchmark [1].
+# LOG_COLORMAP = [disparity error range, RGB color], where the disparity error range is defined as:
+# y = 2^x for x = [-Inf, -4, -3, -2, -1, 0, 1, 2, 3, 4, Inf].
+LOG_COLORMAP = [
```
johnwlambert, line 81: just out of curiosity, does the log colormap look a lot better than a linear colormap?

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,433 @@
...
+    def image_timestamp_list(self) -> Dict[str, List[int]]:
+        """return dict of list of image timestamp (str) for the current log.
+
+        Returns:
+            image_timestamp_list: dictionary of list of image timestamp for all cameras
+        """
+        assert self.image_list is not None
+        assert self._image_list is not None
+
+        if self._image_timestamp_list is None:
```
johnwlambert, line 172: prefer to return early, rather than nest

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,433 @@
...
+        return self._image_timestamp_list[self.current_log]
+
+    @property
+    def timestamp_image_dict(self) -> Dict[str, Dict[int, str]]:
+        """return dict of list of image path (str) for the current log, index by timestamp.
+
+        Returns:
+            timestamp_image_dict: dictionary of list of image path for all cameras, with timestamp as key
+        """
+        if self._timestamp_image_dict is None:
```
johnwlambert, line 190: same note about returning early, rather than nest

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,433 @@
...
+        return self._disparity_list[log_id][name]
+
+    def get_image_at_timestamp(
+        self,
+        timestamp: int,
+        camera: str,
+        log_id: Optional[str] = None,
+        load: bool = True,
+    ) -> Optional[Union[str, np.ndarray]]:
+        """get image or image path at a specific timestamp
```
johnwlambert, line 315: nit: capitalization on "get"

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,433 @@
...
+
+    def __getitem__(self, key: int) -> "ArgoverseStereoLoader":
+        self.counter = key
+        self.current_log = self.log_list[self.counter]
+        self.num_disparity_frame = len(self.disparity_list)
+        self.num_stereo_camera_frame = len(self.image_timestamp_list[STEREO_FRONT_LEFT_RECT])
+        return self
+
+    def get(self, log_id: str) -> "ArgoverseStereoLoader":
+        """get ArgoverseStereoLoader object with current_log set to specified log_id
```
johnwlambert, line 251: nit: capitalization on "get"

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,433 @@
...
+
+        return self._disparity_list[log_id][name]
+
+    def get_image_at_timestamp(
+        self,
+        timestamp: int,
+        camera: str,
+        log_id: Optional[str] = None,
+        load: bool = True,
+    ) -> Optional[Union[str, np.ndarray]]:
```
johnwlambert, line 314: would it be cleaner to have one function that gets the image path, and another function that calls that function to get the image?

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Sure, the function is too long. I've broken it into `eval.py`, `utils.py`, and `constants.py`, similar to the Argoverse detection evaluation.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Changed and put more info.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Changed.

argoverse/evaluation/stereo/disparity_interpolation/disparity_interpolation/source/nn_interpolation.cpp
jhonykaesemodel: Not published on my github yet. Will do it soon.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Sure, made some changes.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Changed. Not sure why black didn't pick this.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Changed.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: It's right. It loops if the list is empty, then fills it and returns the disparity list for the current log_id. Similar to the argoverse_tracking_loader.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: True. Will fix it in the next commit.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Good question. I'm blindly using the KITTI's colormap, which looks pretty good already.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Will fix it in the next commit.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Will fix it in the next commit.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Yeah, will fix in the next commit.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Sure, next commit will return early.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Sure, next commit will return early.

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,438 @@
...
+from argoverse.utils.calibration import Calibration, load_stereo_calib
+from argoverse.utils.camera_stats import RECTIFIED_STEREO_CAMERA_LIST
+
+logger = logging.getLogger(__name__)
+
+STEREO_FRONT_LEFT_RECT = RECTIFIED_STEREO_CAMERA_LIST[0]
+STEREO_FRONT_RIGHT_RECT = RECTIFIED_STEREO_CAMERA_LIST[1]
+
+
+class ArgoverseStereoLoader:
```
johnwlambert, line 21: i do some code repeating here vs the other ArgoverseTrackingLoader. Would it make sense to have both inherit from a parent, and then differ as children?

Which functionality is shared vs. different?

johnwlambert:
Hi Jhony, you will probably need to pull master into your branch, to get Ben's CI fixes

benjaminrwilson (changes requested):

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Yeah, I think there must have better/cleaner ways to do this. I kinda followed the patterns from the argoverse_tracking_loader to be honest.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: good idea.

argoverse/evaluation/stereo/disparity_interpolation/disparity_interpolation/source/nn_interpolation.cpp
jhonykaesemodel: Published now.

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Agreed with Ben, I think it's worth considering a meta class for the evaluation scripts.

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+    rel_error_thresholds: List = [0.1, 0.1, 0.1],
+    figs_fpath: Path = None,
+    save_disparity_error_image: bool = False,
+):
+    """Compute the disparity error metrics."""
+    pred_disparity = cv2.imread(str(pred_fpath), cv2.IMREAD_ANYCOLOR | cv2.IMREAD_ANYDEPTH)
+    gt_disparity = cv2.imread(str(gt_fpath), cv2.IMREAD_ANYCOLOR | cv2.IMREAD_ANYDEPTH)
+    gt_obj_disparity = cv2.imread(str(gt_obj_fpath), cv2.IMREAD_ANYCOLOR | cv2.IMREAD_ANYDEPTH)
+
+    pred_disparity = np.float32(pred_disparity) / 256.0
```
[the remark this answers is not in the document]
jhonykaesemodel: how about `DISPARITY_NORM = 2.0 ** 8`  # 256.0 ?

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    abs_error_thresholds: List[int] = DEFAULT_ABS_ERROR_THRESHOLDS,
+    rel_error_thresholds: List[float] = DEFAULT_REL_ERROR_THRESHOLDS,
+    save_disparity_error_image: bool = False,
+) -> pd.DataFrame:
+    """Compute the disparity error metrics."""
+    pred_disparity = cv2.imread(str(pred_fpath), cv2.IMREAD_ANYCOLOR | cv2.IMREAD_ANYDEPTH)
+    gt_disparity = cv2.imread(str(gt_fpath), cv2.IMREAD_ANYCOLOR | cv2.IMREAD_ANYDEPTH)
+    gt_obj_disparity = cv2.imread(str(gt_obj_fpath), cv2.IMREAD_ANYCOLOR | cv2.IMREAD_ANYDEPTH)
+
+    pred_disparity = np.float32(pred_disparity) / 256.0
```
johnwlambert, line 34: are these stored as uint8? i thought we were saving them as uint16

tests/test_data/stereo/rectified_stereo_images_v1.1/test/1/vehicle_calibration_stereo_info.json
```diff
@@ -0,0 +1 @@
...ehicle_SE3_camera_": {"rotation": {"coefficients": [0.9991537960018827, -0.002823111217897565, -0.04092474315547999, 0.0029811701960101686]}, "translation": [0, 0, 0]}}}, {"key": "image_raw_stereo_front_left_rect_refined", "value": {"focal_length_x_px_": 3894.034002486376, "focal_length_y_px_": 3894.034002486376, "focal_center_x_px_": 937.5700225830078, "focal_center_y_px_": 1063.1990280151367, "Txf_": 0.0, "skew_": 0.0, "distortion_coefficients_": [-0.09737880168798231, 0.025226817408578506, 0.6478859260574071], "vehicle_SE3_camera_": {"rotation": {"coefficients": [0.5221778964049267, -0.5151654646934947, 0.48107093840642473, -0.48010992565663874]}, "translation": [1.7241920199697882, 0.0810346048840847, 1.3553077959056061]}}}], "vehicle_SE3_up_lidar_": {"rotation": {"coefficients": [0.9998848409086823, 0.0011389828365580806, -0.0010769852622658116, -0.015094626257773826]}, "translation": [1.3504724404154866, -0.003966310091761113, 1.5700506059693788]}, "vehicle_SE3_down_lidar_": {"rotation": {"coefficients": [5.6898203165648547e-05, -0.994250135333888, -0.10708208174631563, -0.0003048278454676608]}, "translation": [1.3502875642048289, -0.0019957763317242708, 1.4771699630675978]}}
```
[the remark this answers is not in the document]
jhonykaesemodel: Sure!

tests/test_eval_stereo.py
```diff
@@ -0,0 +1,49 @@
...
+
+import math
+from pathlib import Path
+
+from argoverse.evaluation.stereo.eval import StereoEvaluator
+
+_ROOT = Path(__file__).resolve().parent
+
+
+def test_stereo_evaluation() -> None:
```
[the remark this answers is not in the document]
jhonykaesemodel: yeap

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: These are inverse depth (disparity in pixel units), so uint8 is enough. We can always recover true depth with great precision.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Agreed, definitely a meta data loader would be helpful here. I'd have to carefully think about it to not break things in the forecast and tracking loaders though. The problem is that we decided to have a different data structure for the stereo dataset. We have 2 folders: rectified_stereo_images_v1.1 and disparity_maps_v1.1. Also I'm using the RECTIFIED_STEREO_CAMERA_LIST. So the methods need to be a bit different from the tracking/forecasting loader.

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: Fixing my previous comment: the disparity maps are saved as uint16 PNG images. The true disparity value for each pixel is computed by converting the uint16 value to float and dividing it by 256.

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,440 @@
...
+                                    self.root_dir, "rectified_stereo_images_v1.1", self.split_name, log, camera, "*.jpg"
+                                )
+                            )
+                        )
+                    )
+                    self.image_count += len(self._image_list[log][camera])
+        return self._image_list[self.current_log]
+
+    @property
+    def disparity_list(self) -> Dict[str, List[str]]:
```
johnwlambert, line 141: nit: capitalize Return

argoverse/utils/camera_stats.py
johnwlambert: Just to clarify -- are we releasing new rectified imagery?

argoverse/evaluation/stereo/disparity_interpolation/disparity_interpolation/source/nn_interpolation.cpp
johnwlambert: Cool. Maybe you could add a high-level algorithm overview to that other repo, and a few small unit tests on toy data too?

jhonykaesemodel: Definitely, already in my to-do list.

argoverse/utils/camera_stats.py
jhonykaesemodel: Yes, we released rectified imagery.

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+        gt_root_fpath: Path,
+        figs_fpath: Path,
+        abs_error_thresholds: List = [10, 5, 3],
+        rel_error_thresholds: List = [0.1, 0.1, 0.1],
+        save_disparity_error_image: bool = False,
+        num_procs: int = -1,
+    ) -> None:
+        """
+        Args:
+            pred_fpath_root: Path to the folder which contains the stereo predictions.
```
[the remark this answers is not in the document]
jhonykaesemodel: Oh yeah, forgot about them.

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,348 @@
...
+                    self.save_disparity_error_image,
+                )
+                for (pred_fpath, gt_fpath, gt_obj_fpath) in zip(pred_fpaths, gt_fpaths, gt_obj_fpaths)
+            ]
+            with Pool(self.num_procs) as p:
+                errors = p.starmap(compute_disparity_error, args)
+
+        data = pd.concat(errors)
+        data_sum = data.sum()
+        summary: dict() = dict()
```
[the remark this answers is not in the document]
jhonykaesemodel: Good call! Changed it.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: yeah, just fixed all docstrings that were initialized with low caps.

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    gt_disparity = np.float32(gt_disparity) / 256.0
```
benjaminrwilson, line 35: I might consider making a constant for `256.0`: `RGB_NORMALIZATION: int = 2 ** 8`

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+    for abs_error_thresh, rel_error_thresh in zip(abs_error_thresholds, rel_error_thresholds):
+        bad_pixels = (abs_err > abs_error_thresh) & (rel_err > rel_error_thresh)
+
+        errors[f"num_errors_bg:{abs_error_thresh}"] = np.sum(bg_mask & bad_pixels)
+        errors[f"num_errors_fg:{abs_error_thresh}"] = np.sum(fg_mask & bad_pixels)
+        errors[f"num_errors_bg_est:{abs_error_thresh}"] = np.sum(bg_mask & pred_mask & bad_pixels)
+        errors[f"num_errors_fg_est:{abs_error_thresh}"] = np.sum(fg_mask & pred_mask & bad_pixels)
+
+    if save_disparity_error_image:
```
benjaminrwilson, line 72: Not pressing, but it might be nice to have the image saving functionality in a separate function.

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+    for abs_error_thresh in abs_error_thresholds:
+        columns += [
+            f"num_errors_bg:{abs_error_thresh}",
+            f"num_errors_fg:{abs_error_thresh}",
+            f"num_errors_bg_est:{abs_error_thresh}",
+            f"num_errors_fg_est:{abs_error_thresh}",
+        ]
+
+    return pd.DataFrame([[0] * num_fields], columns=columns)
```
benjaminrwilson, line 150: Maybe a comment explaining the first arg of the dataframe? I'm not sure what that's doing.

argoverse/utils/calibration.py
```diff
@@ -235,6 +236,34 @@ def load_calib(calib_filepath: Union[str, Path]) -> Dict[Any, Calibration]:
     return calib_list
 
 
+def load_stereo_calib(calib_filepath: Union[str, Path]) -> Dict[Any, Calibration]:
```
benjaminrwilson, line 239: Could we change `calib_filepath: Union[str, Path]` -> `calib_filepath: _PathLike` for consistency? See `map_api.py` for an example. Eventually, we should move the typing into one place rather than copying type aliases everywhere (as seen in `map_api.py`).

tests/test_argoverse_stereo_loader.py
```diff
@@ -0,0 +1,68 @@
...
+from argoverse.data_loading.argoverse_stereo_loader import ArgoverseStereoLoader
+from argoverse.utils.camera_stats import RECTIFIED_STEREO_CAMERA_LIST
+
+TEST_DATA_LOC = str(pathlib.Path(__file__).parent.parent / "tests" / "test_data" / "stereo")
+
+STEREO_DISPARITY_LIST = ["stereo_front_left_rect_disparity", "stereo_front_left_rect_objects_disparity"]
+
+
+@pytest.fixture  # type: ignore
+def data_loader() -> ArgoverseStereoLoader:
```
benjaminrwilson, line 17: Let's add docstrings to these functions to describe what they're testing at some point. Not pressing IMO.

benjaminrwilson (changes requested):

argoverse/evaluation/stereo/utils.py
benjaminrwilson: Correction: `RGB_NORMALIZATION: float = 2.0 ** 8`

benjaminrwilson: Well, whatever you think is best. I guess `int` makes sense as long as it doesn't interfere with any casting.

jhonykaesemodel: Yeap, changed to `DISPARITY_NORMALIZATION: float = 2.0 ** 8`

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def interpolate_disparity(disp: np.ndarray) -> np.ndarray:
+    """Interpolate disparity image to inpaint holes.
+    The expected run time for the Argoverse stereo image with 2056 × 2464 pixels is ~50 ms.
```
johnwlambert, line 155: might be helpful to add the expected input and output sizes/types here of `disp` and the return arg. are they (M,N) float or uint8 or uint16 matrices?

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,207 @@
...
+                    self.rel_error_thresholds,
+                    self.save_disparity_error_image,
+                )
+                for (pred_fpath, gt_fpath, gt_obj_fpath) in zip(pred_fpaths, gt_fpaths, gt_obj_fpaths)
+            ]
+            with Pool(self.num_procs) as p:
+                errors = p.starmap(compute_disparity_error, args)
+
+        data = pd.concat(errors)
+        data_sum = data.sum()
```
johnwlambert, line 150: maybe an inline comment -- is this summing along rows or columns? does that sum over instances or pixels etc?

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: Agreed!

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,207 @@
...
+        save_disparity_error_image: bool = False,
+        num_procs: int = -1,
+    ) -> None:
+        """
+        Args:
+            pred_fpath_root: Path to the folder which contains the stereo predictions.
+            gt_fpath_root: Path to the folder which contains the stereo ground truth.
+            figs_fpath: Path to the folder which will contain the output figures.
+            abs_error_thresholds: List with the absolute disparity error thresholds.
+            rel_error_thresholds: List with the relative disparity error thresholds.
```
johnwlambert, line 89: i think the List part is already clear from the function signature/type hint, so I would say something a bit more descriptive like
```python
abs_error_thresholds: Absolute disparity error thresholds, in pixels
```

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,207 @@
...
+        all:10,
+        fg:10,
+        bg:10,
+        all*:10,
+        fg*:10,
+        bg*:10,
+        all:5,
+        fg:5,
+        bg:5,
+        all*:5,
```
johnwlambert, line 42: what does the asterisk represent?

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,440 @@
...
+        """Return list of all image paths (str) for all cameras for the current log
+
+        Returns:
+            image_list: dictionary of list of image, with camera name as key
+        """
+        if self._disparity_list is None:
```
johnwlambert, line 147: would be good to remove the extra nesting here, and just return early

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # Blend the log-colormap image to the top of the disparity error image.
+        c_img = colormap_image[:, :, ::-1]
+        d_img = disparity_error_image[:, :, ::-1]
+        x_offset = 30
+        y_offset = 30
+        d_img[y_offset : y_offset + c_img.shape[0], x_offset : x_offset + c_img.shape[1]] = c_img
+
+        # Save the blended disparity error image to a PNG file.
+        log_id = str(gt_fpath).split("/")[-3]
```
johnwlambert, line 122: perhaps simpler as `Path(gt_fpath).parts[-3]`

tests/test_argoverse_stereo_loader.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def test_loading_image_disparity(data_loader: ArgoverseStereoLoader) -> None:
+    camera = RECTIFIED_STEREO_CAMERA_LIST[0]
+    disparity = STEREO_DISPARITY_LIST[0]
+    disparity_objects = STEREO_DISPARITY_LIST[1]
+    log = "1"
+    image_1_at_timestamp = data_loader.get_image_at_timestamp(1, camera, log)
+    image_1 = data_loader.get_image(0, camera, log)
+    disparity_1 = data_loader.get_disparity_map(0, disparity, log)
```
johnwlambert, line 64: keyword arg might help explain what `0` represents a bit more:
```python
disparity_1 = data_loader.get_disparity_map(idx=0, name=disparity, log_id=log)
```

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: Yeah, I put a comment. It initializes the accumulator with zeroes in all fields.

argoverse/evaluation/stereo/utils.py
johnwlambert: Cool. Do you mind adding as an inline comment here that we are reading uint16 files?

What is the resolution and min/max values we can represent without uint16/256 representation? maybe useful to document somewhere

argoverse/utils/camera_stats.py
johnwlambert: Very cool

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,160 @@
...
+        save_dir = f"{figs_fpath}/{log_id}/"
+        Path(save_dir).mkdir(parents=True, exist_ok=True)
+        cv2.imwrite(f"{save_dir}/disparity_error_{timestamp}.png", d_img)
+
+    return errors
+
+
+def accumulate_stereo_metrics(abs_error_thresholds: List[int]) -> pd.DataFrame:
+    """Metrics accumulator."""
+    num_fields = 4 + 4 * len(abs_error_thresholds)
```
johnwlambert, line 133: maybe 4 can be a named constant? perhaps `NUM_REL_THRESHOLDS` or something? or what does it represent?

argoverse/evaluation/stereo/eval.py
johnwlambert: oh nevermind, i see the explanation above

argoverse/evaluation/stereo/constants.py
```diff
@@ -0,0 +1,31 @@
+from typing import List
+
+import numpy as np
+
+# We consider the disparity of a pixel to be correctly estimated if the absolute disparity error is less than a
+# threshold and its relative error is less than 10% of its true value.
+# We define three disparity error thresholds: 3, 5, and 10 pixels.
+# Similar to the KITTI Stereo 2015 evaluation, we empirically found that the combination of absolute and relative
+# disparity errors ensures an evaluation which is faithful with respect to the annotation errors in the ground truth.
```
johnwlambert, line 9: "with respect to the annotation errors in the ground truth." --> this is a bit confusing to me since we didn't hand-annotate the GT, but it's estimated from the scene flow. So I would say "with respect to the errors in the ground truth."

argoverse/utils/calibration.py
jhonykaesemodel: I think we can do that in another PR as I've seen that typing style in other functions too.

tests/test_argoverse_stereo_loader.py
jhonykaesemodel: yeah, agreed. Many of our unit tests have no docstrings.

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: Sure!

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Put a comment.

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: This is cool! much cleaner.

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: agreed, put new constants, `NUM_EVAL_PIXEL_REGIONS` and  `NUM_EVAL_ERROR_REGIONS`, in the `constants.py`

argoverse/evaluation/stereo/constants.py
jhonykaesemodel: Agreed, thanks for pointing this out.

johnwlambert:
Screen Shot 2021-04-15 at 8 09 41 PM
 2 small typos in the notebook here

johnwlambert:
> My bad, is must be `if self._calib is not None:`... It passed all tests somehow. Do you think we should use the `pass` since the function returns nothing? I think the older format is cleaner.

Would a simple `return` instead of `pass` accomplish the logic?

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,208 @@
...
+                    self.rel_error_thresholds,
+                    self.save_disparity_error_image,
+                )
+                for (pred_fpath, gt_fpath, gt_obj_fpath) in zip(pred_fpaths, gt_fpaths, gt_obj_fpaths)
+            ]
+            with Pool(self.num_procs) as p:
+                errors = p.starmap(compute_disparity_error, args)
+
+        data = pd.concat(errors)
+        # Sums over the instances along the rows containing the disparity error metrics.
```
johnwlambert, line 150: Thanks for the updates, Jhony. I believe the columns are the disparity error metrics, so it might be clearer to say
```
# Sums over all frames (row dimension) for each disparity error metric
```

tests/test_eval_stereo.py
```diff
@@ -0,0 +1,211 @@
...
+                         [150.0, 120.0, 190.0]], dtype=np.float32)
+
+        gt = np.array([[  1.0,   5.0,  10.0],
+                       [ 50.0,  70.0,  90.0],
+                       [100.0, 150.0, 200.0]], dtype=np.float32)
+
+        gt_obj = np.array([[0.0, 0.0, 0.0],
+                           [0.0, 0.0, 0.0],
+                           [0.0, 0.0, 0.0]], dtype=np.float32)
+
```
johnwlambert, line 170: this looks great. maybe you could add like 5 more lines to the docstring just briefly showing how we arrive at the final metrics, based on these 3 dummy images?

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: Right!

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,459 @@
...
+
+    def get_log_stereo_img_paths(self, camera: str, log_id: Optional[str] = None) -> List[str]:
+        """Get list of stereo image paths for specific log
+
+        Args:
+            log_id: log_id, if not specified will use self.current_log
+            camera: camera based on camera_stats.RECTIFIED_STEREO_CAMERA_LIST
+
+        Returns:
+            List[str]: list of stereo image paths for the log (str or List[str]),
```
johnwlambert, line 318: no need to repeat the type here since it's already self-evident from the function signature, so we can just say "List of stereo image paths for the log"

When would it be str instead of List[str]?

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,459 @@
...
+            x
+            for x in os.listdir(os.path.join(self.root_dir, "rectified_stereo_images_v1.1", self.split_name))
+            if valid_log(x)
+        ]
+
+        return self._log_list
+
+    @property
+    def image_list(self) -> Dict[str, List[str]]:
+        """Return list of all image path (str) for the current log
```
johnwlambert, line 122: are these ring camera images, or just stereo images? Are they rectified or unrectified?

Maybe
```
"Return list of all rectified stereo image paths (str) for the current log"
```

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,459 @@
...
+                "rectified_stereo_images_v1.1",
+                self.split_name,
+                log,
+                "vehicle_calibration_stereo_info.json",
+            )
+            self._calib[log] = load_stereo_calib(calib_filename)
+
+    @property
+    def log_list(self) -> List[str]:
+        """Return list of log (str) in the current dataset directory
```
johnwlambert, line 93: typo: "Return list of log_ids in the current dataset directory"

The type is already self-evident from the function signature, I believe

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,459 @@
...
+            calib: Calibration object for the current log
+        """
+        self.__ensure_calib_is_populated()
+        assert self._calib is not None
+        return self._calib[self.current_log]
+
+    def __ensure_calib_is_populated(self) -> None:
+        """Load up calibration object for all logs
+
+        Returns:
```
johnwlambert, line 74: Returns section is not needed when None is returned (can actually make less readable)

tests/test_eval_stereo.py
jhonykaesemodel: will do

argoverse/data_loading/argoverse_stereo_loader.py
```diff
@@ -0,0 +1,459 @@
...
+
+DISPARITY_NORMALIZATION: float = 2.0 ** 8  # 256.0
+
+
+class ArgoverseStereoLoader:
+    def __init__(self, root_dir: str, split_name: str) -> None:
+        # initialize class member
+        self.RECTIFIED_STEREO_CAMERA_LIST = RECTIFIED_STEREO_CAMERA_LIST
+        self._log_list: Optional[List[str]] = None
+        self._image_list: Optional[Dict[str, Dict[str, List[str]]]] = None
```
johnwlambert, line 30: inline comments here might help, e.g.
```python
# for each log, provides image paths for each rectified stereo camera
self._image_list: Optional[Dict[str, Dict[str, List[str]]]] = None
```
It does seem like we are using a dictionary as a class, whereas NamedTuple is preferred for that. But at least inline comments would be super helpful.

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,252 @@
...
+            f"num_errors_fg:{abs_error_thresh}",
+            f"num_errors_bg_est:{abs_error_thresh}",
+            f"num_errors_fg_est:{abs_error_thresh}",
+        ]
+
+    return pd.DataFrame([[0] * num_fields], columns=columns)
+
+
+def interpolate_disparity(disp: np.ndarray) -> np.ndarray:
+    """Interpolate disparity image to inpaint holes.
```
johnwlambert, line 132: might be helpful to add 5-10 words just explaining is it "nearest neighbor" interpolation, or "linear", or "bilinear"? is there a URL or reference we could add to the docstring detailing the algorithm?

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Got it. It's supposed to be List[str] only. Will fix.

argoverse/data_loading/argoverse_stereo_loader.py
jhonykaesemodel: Makes sense. It's rectified stereo images.

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,208 @@
...
+            d1_all = (data_sum[f"num_errors_bg:{abs_error_thresh}"] + data_sum[f"num_errors_fg:{abs_error_thresh}"]) / (
+                data_sum["num_pixels_bg"] + data_sum["num_pixels_fg"]
+            )
+            d1_fg = data_sum[f"num_errors_fg:{abs_error_thresh}"] / data_sum["num_pixels_fg"]
+            d1_bg = data_sum[f"num_errors_bg:{abs_error_thresh}"] / data_sum["num_pixels_bg"]
+            d1_all_est = (
+                data_sum[f"num_errors_bg_est:{abs_error_thresh}"] + data_sum[f"num_errors_fg_est:{abs_error_thresh}"]
+            ) / (data_sum["num_pixels_bg_est"] + data_sum["num_pixels_fg_est"])
+            d1_fg_est = data_sum[f"num_errors_fg_est:{abs_error_thresh}"] / data_sum["num_pixels_fg_est"]
+            d1_bg_est = data_sum[f"num_errors_bg_est:{abs_error_thresh}"] / data_sum["num_pixels_bg_est"]
```
johnwlambert, line 164: what does `d1` mean here?

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: sure

argoverse/evaluation/stereo/eval.py
jhonykaesemodel: it was our old notation, following kitti. I'll remove those.

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,252 @@
...
+    pred_fpath: Path,
+    gt_fpath: Path,
+    gt_obj_fpath: Path,
+    figs_fpath: Path,
+    abs_error_thresholds: List[int] = DEFAULT_ABS_ERROR_THRESHOLDS,
+    rel_error_thresholds: List[float] = DEFAULT_REL_ERROR_THRESHOLDS,
+    save_disparity_error_image: bool = False,
+) -> pd.DataFrame:
+    """
+    Compute the disparity errors.
```
johnwlambert, line 33: should go on the same line above
```
) -> pd.DataFrame:
    """Compute the disparity errors.

        Args:
```

https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,252 @@
...
+
+    Args:
+        pred_fpath: Path to the folder which contains the stereo predictions.
```
johnwlambert, line 36: path to the folder, or path to the file?

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,252 @@
...
+        gt_fpath: Path to the folder which contains the stereo ground truth.
+        gt_obj_fpath: Path to the folder which contains the stereo ground truth for foreground objects.
```
johnwlambert, line 38: path to file?

argoverse/evaluation/stereo/eval.py
```diff
@@ -0,0 +1,208 @@
...
+    a simple nearest neighbor interpolation as in the KITTI Stereo 2015 benchmark [1] to assure we can compare it to
+    our semi-dense ground-truth disparity map. Current deep stereo matching methods normally predict disparity maps
+    with 100% density. Thus, an interpolation step is not needed for the evaluation.
+
+    The disparity errors metrics are the following:
+
+    1. all: Percentage of stereo disparity errors averaged over all ground truth pixels in the reference frame
+            (left stereo image).
+    2. bg: Percentage of stereo disparity errors averaged only over background regions.
+    3. fg: Percentage of stereo disparity errors averaged only over foreground regions.
```
johnwlambert, line 21: maybe you could add one line explaining about how you computed the foreground region boundary?

argoverse/evaluation/stereo/utils.py
```diff
@@ -0,0 +1,252 @@
...
+    pred_mask = pred_disparity > 0
+    gt_mask = gt_disparity > 0
+    obj_mask = gt_obj_disparity > 0
+    bg_mask = np.logical_and(gt_mask, np.logical_not(obj_mask))
+    fg_mask = np.logical_and(gt_mask, obj_mask)
+
+    # If the density of the predicted disparity is less than 100%, then interpolate to fill up the holes.
+    num_pixels_all = pred_disparity.size
+    num_pixels_all_est = np.sum(pred_mask)
+    density = num_pixels_all_est / max(num_pixels_all, 1.0)
```
johnwlambert, line 68: why do we need the `max(x, 1.0)` here?

argoverse/data_loading/stereo_dataloader.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        disparity_map_fpaths = sorted(
+            glob.glob(f"{self.data_dir}/disparity_maps_v1.1/{self.split_name}/{log_id}/{disparity_name}/*.png")
+        )
+
+        return disparity_map_fpaths
+
+    def get_rectified_stereo_image(self, stereo_img_path: str) -> np.ndarray:
+        """
+        Get the rectified stereo image.
```
johnwlambert, line 82: Should be `""" Get the rectified stereo image.` (on the same line as triple quote)

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: file. Fixed the docstring.

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: yes, fixed.

argoverse/evaluation/stereo/utils.py
jhonykaesemodel: we actually don't need this.

johnwlambert (approved):

PR #209, opened 2021-04-19
Added brier score for p-metrics in forecasting

jagjeet-singh:
This PR adds brier score based metric to forecasting. The two new metrics are:
`brier-minFDE = (1.0-p)^2 + minFDE` and `brier-minADE = (1.0-p)^2 + minADE`

Motivation for using brier score for assessing uncertainty:
- Brier score is a commonly used metric in probability calibration ("Brier, G. W. Verification of forecasts expressed in terms of probability. Monthly weather review, 1950.")
- The contribution is in [0.0, 1.0], thereby making sure probabilities are not penalized too high.
- No tunable parameters.
- Given that both minFDE and f(p) are in a similar range, the methods will be motivated to do well in both. 

image

Top methods from eval ai leaderboard:
image

### DIFF
--- a/argoverse/evaluation/eval_forecasting.py
+++ b/argoverse/evaluation/eval_forecasting.py
@@ -79,11 +79,11 @@ def get_displacement_errors_and_miss_rate(
         forecasted_probabilities: Probabilites associated with forecasted trajectories.
 
     Returns:
-        metric_results: Metric values for minADE, minFDE, MR, p-minADE, p-minFDE, p-MR
+        metric_results: Metric values for minADE, minFDE, MR, p-minADE, p-minFDE, p-MR, brier-minADE, brier-minFDE
     """
     metric_results: Dict[str, float] = {}
-    min_ade, prob_min_ade = [], []
-    min_fde, prob_min_fde = [], []
+    min_ade, prob_min_ade, brier_min_ade = [], [], []
+    min_fde, prob_min_fde, brier_min_fde = [], [], []
     n_misses, prob_n_misses = [], []
     for k, v in gt_trajectories.items():
         curr_min_ade = float("inf")
@@ -122,6 +122,7 @@ def get_displacement_errors_and_miss_rate(
                 )
                 + curr_min_ade
             )
+            brier_min_ade.append((1 - pruned_probabilities[min_idx]) ** 2 + curr_min_ade)
             prob_min_fde.append(
                 min(
                     -np.log(pruned_probabilities[min_idx]),
@@ -129,6 +130,8 @@ def get_displacement_errors_and_miss_rate(
                 )
                 + curr_min_fde
             )
+            brier_min_fde.append((1 - pruned_probabilities[min_idx]) ** 2 + curr_min_fde)
+
     metric_results["minADE"] = sum(min_ade) / len(min_ade)
     metric_results["minFDE"] = sum(min_fde) / len(min_fde)
     metric_results["MR"] = sum(n_misses) / len(n_misses)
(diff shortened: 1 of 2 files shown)


johnwlambert:
Thanks for the PR, @jagjeet-singh. Would you mind adding a few sentences here explaining the motivation for the change, and links to any relative literature? Maybe those could go into the docstrings, as well?

Which scenarios do we have tested so far?

integration_tests/test_eval_forecasting.py
```diff
@@ -169,25 +184,130 @@ def test_compute_forecasting_metric() -> None:
...
+        "minFDE": 1.0,
+        "DAC": 1.0,
+        "MR": 0.0,
+        "p-minADE": 1.16,
+        "p-minFDE": 1.16,
+        "p-MR": 0.4,
+    }
+    assert_metrics_almost_equal(expected_metrics, metrics)
+
+    # Case 5: Top-2, 2 forecast, uniform probabilities
```
johnwlambert, line 279: probably would be cleaner to make this 5 separate tests, with a pytest fixture?

integration_tests/test_eval_forecasting.py
```diff
@@ -169,25 +184,130 @@ def test_compute_forecasting_metric() -> None:
...
+        miss_threshold,
+        forecasted_probabilities,
+    )
+
+    expected_metrics: Dict[str, float] = {
+        "minADE": 1.0,
+        "minFDE": 1.0,
+        "DAC": 1.0,
+        "MR": 0.0,
+        "p-minADE": 1.16,
```
johnwlambert, line 273: these numbers seem a bit hard to work out by hand -- can we come up with a simple example, e.g. straight lines or something, where we can work it out by hand?

argoverse/evaluation/eval_forecasting.py
```diff
@@ -115,20 +115,9 @@ def get_displacement_errors_and_miss_rate(
...
-                + curr_min_ade
-            )
-            prob_min_fde.append(
-                min(
-                    -np.log(pruned_probabilities[min_idx]),
-                    -np.log(LOW_PROB_THRESHOLD_FOR_METRICS),
-                )
-                + curr_min_fde
-            )
+            prob_min_ade.append((1 - pruned_probabilities[min_idx]) ** 2 + curr_min_ade)
```
johnwlambert, line 118: what was the rationale for using -log before?

maybe a reference to papers that use Brier would be useful here

integration_tests/test_eval_forecasting.py
jagjeet-singh: I guess the final metric values need not be easy to work out by hand as long as the test example is self-explanatory. The test example covers a few cases which straight-line examples will miss.

integration_tests/test_eval_forecasting.py
jagjeet-singh: I was on the fence about it but ended up refactoring the tests anyway. It's more organized now.

argoverse/evaluation/eval_forecasting.py
jagjeet-singh: Added a reference to the original brier score in the PR description. To my knowledge, no one has used brier score with discplacent error. So there's not reference for that.
Brier score is a commonly used metric for probability calibration.

integration_tests/test_eval_forecasting.py
jagjeet-singh: Thanks for the push.

integration_tests/test_eval_forecasting.py
jagjeet-singh: Actually, this particular case can even be worked out by hand. The forecast and ground truth are 1.0m apart in each frame. This is not any more complicated than straight lines.
The other case where the forecast is a turning trajectory might not be that straightforward.

PR #211, opened 2021-04-23
Move some integration tests into additional unit tests

johnwlambert:
We are now using the map files in CI for the detection/tracking eval, so we can use it for forecasting unit tests as well, in CI.


argoverse/evaluation/eval_forecasting.py
```diff
@@ -67,6 +67,9 @@ def get_displacement_errors_and_miss_rate(
     """Compute min fde and ade for each sample.
 
     Note: Both min_fde and min_ade values correspond to the trajectory which has minimum fde.
+    The Brier Score is defined here:
```
jagjeet-singh, line 70: Thanks for adding it here. I missed it in my PR.

jagjeet-singh (approved):

argoverse/evaluation/eval_forecasting.py
johnwlambert: No problem

PR #212, opened 2021-04-23
Add factory functions for Sim(2) clasmethod and calibration helpers in loader

johnwlambert:
- Class methods to load Sim(2) from disk or numpy matrix
- Add additional helpers to `SimpleArgoverseTrackingLoader` to reduce boilerplate on user side


argoverse/utils/sim2.py
```diff
@@ -108,3 +109,22 @@ def transform_from(self, point_cloud: np.ndarray) -> np.ndarray:
     def transform_point_cloud(self, point_cloud: np.ndarray) -> np.ndarray:
         """Alias for `transform_from()`, for synchrony w/ API provided by SE(2) and SE(3) classes."""
         return self.transform_from(point_cloud)
+
+    @classmethod
+    def from_json(cls, json_fpath: str) -> "Sim2":
```
benjaminrwilson, line 114: What do you think about a `Union[_Pathlike, str]` here?

benjaminrwilson: At some point we should define `Union[_Pathlike, str]` in a `typing` module instead of copy-pasting throughout the library IMO.

argoverse/utils/sim2.py
```diff
@@ -108,3 +109,22 @@ def transform_from(self, point_cloud: np.ndarray) -> np.ndarray:
...
+        """Generate class inst. from a JSON file containing Sim(2) parameters as flattened matrices (row-major)."""
```
benjaminrwilson, line 115: Nice 👍 . I appreciate the row-major comment.

benjaminrwilson (changes requested):

argoverse/utils/sim2.py
johnwlambert: Good point about the PathLike -- I just updated it

argoverse/utils/sim2.py
```diff
@@ -108,3 +110,22 @@ def transform_from(self, point_cloud: np.ndarray) -> np.ndarray:
...
+        t = np.array(json_data["t"]).reshape(2)
+        s = float(json_data["s"])
+        return cls(R, t, s)
+
+    @classmethod
+    def from_matrix(cls, T: np.ndarray) -> "Sim2":
+        """Generate class instance from a 3x3 Numpy matrix """
+        R = T[:2, :2]
+        t = T[:2, 2]
+        s = 1 / T[2, 2]
```
benjaminrwilson, line 130: We should verify the user isn't passing a scale of zero here too.

tests/test_simple_track_dataloader.py
```diff
@@ -27,6 +30,29 @@ def test_get_log_calibration_data(
...
+    # check intrinsics, should be 3x4 since we use 4x4 extrinsics
+    expected_K = np.array([[1392.11, 0, 980.18, 0],[0, 1392.11, 604.35, 0],[0, 0, 1, 0]])
+    assert np.allclose(expected_K, cam_config.intrinsic, atol=0.01)
+    assert cam_config.distortion_coeffs == [-0.1720396447593493, 0.11689572230654095, -0.02511932396889168]
+
+    # check extrinsics
+    qw, qx, qy, qz = 0.49605542988442836, -0.49896196582115804, 0.5027901707576079, -0.5021633313331392
+    R = Rotation.from_quat([qx,qy,qz,qw]).as_matrix()
+    t = [1.6519358245144808, -0.0005354981581146487, 1.3613890006792675]
+    egoTc = np.eye(4)
```
benjaminrwilson, line 51: I know these are probably GTSAM conventions, but do you think we could find a snake case equivalent?

argoverse/utils/sim2.py
johnwlambert: thanks -- I added an assert and a new unit test that checks to see if an exception is raised for the degenerate case

argoverse/utils/sim2.py
```diff
@@ -108,3 +110,25 @@ def transform_from(self, point_cloud: np.ndarray) -> np.ndarray:
...
+        with open(json_fpath, "r") as f:
+            json_data = json.load(f)
+
+        R = np.array(json_data["R"]).reshape(2, 2)
+        t = np.array(json_data["t"]).reshape(2)
+        s = float(json_data["s"])
+        return cls(R, t, s)
+
+    @classmethod
+    def from_matrix(cls, T: np.ndarray) -> "Sim2":
```
benjaminrwilson, line 126: How do we handle return typing when a method can raise an exception? I'm not sure what the correct convention is.

johnwlambert:
@tagarwal-argoai, if you have a few minutes, would you mind taking a brief look at this also? Thanks very much.

PR #218, opened 2021-05-06
make constants explicit in synchronization database

johnwlambert:
Make the frame rate constants more readable/ clear.

### DIFF
--- a/argoverse/data_loading/synchronization_database.py
+++ b/argoverse/data_loading/synchronization_database.py
@@ -9,8 +9,12 @@ from typing import Dict, Iterable, List, Optional, Tuple, cast
 
 import numpy as np
 
+import argoverse.utils.metric_time.TimeUnit.Millisecond as Millisecond
+import argoverse.utils.metric_time.TimeUnit.Nanosecond as Nanosecond
+import argoverse.utils.metric_time.TimeUnit.Second as Second
 from argoverse.utils.camera_stats import RING_CAMERA_LIST, STEREO_CAMERA_LIST
 from argoverse.utils.json_utils import read_json_file
+from argoverse.utils.metric_time import to_metric_time
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


argoverse/data_loading/synchronization_database.py
```diff
@@ -14,6 +14,17 @@
 
 logger = logging.getLogger(__name__)
 
+RING_CAMERA_FPS = 30
+STEREO_CAMERA_FPS = 5
+LIDAR_FRAME_RATE_HZ = 10
+
+# in milliseconds
+RING_CAMERA_SHUTTER_INTERVAL_MS = (1 / RING_CAMERA_FPS) * 1000 # evaluates to 33.3 ms
```
benjaminrwilson: I think that 1000 might be more readable as `1e3`.

benjaminrwilson: (in terms of quickly determining the units conversion)

argoverse/data_loading/synchronization_database.py
```diff
@@ -14,6 +14,17 @@
 
 logger = logging.getLogger(__name__)
 
+RING_CAMERA_FPS = 30
```
benjaminrwilson, line 17: Can we add typing here?

argoverse/data_loading/synchronization_database.py
```diff
@@ -59,21 +70,20 @@ def find_closest_integer_in_ref_arr(query_int: int, ref_arr: np.ndarray) -> Tupl
...
-    # LiDAR is 10 Hz
-    # Max we are halfway between 33.3 milliseconds on either side
+    # Max difference between camera and LiDAR observation would be if the LiDAR timestamp is halfway between
+    # two camera observations (i.e. RING_CAMERA_SHUTTER_INTERVAL_MS / 2 milliseconds on either side)
     # then convert milliseconds to nanoseconds
-    MAX_LIDAR_RING_CAM_TIMESTAMP_DIFF = 33.3 * (1.0 / 2) * (1.0 / 1000) * 1e9
-    # Stereo Camera is 5 Hz (once per 200 milliseconds)
-    # LiDAR is 10 Hz
-    # Since Stereo is more sparse, we look for 200 millisecond on either side
+    MAX_LIDAR_RING_CAM_TIMESTAMP_DIFF = RING_CAMERA_SHUTTER_INTERVAL_MS * (1.0 / 2) * (1.0 / 1000) * 1e9
```
benjaminrwilson, line 77: I think we should consider making this more readable. One idea:

`MAX_LIDAR_RING_CAM_TIMESTAMP_DIFF = 0.5 * (1e9 / 1e3) * RING_CAMERA_SHUTTER_INTERVAL_MS`

I think ideally we would have a function to change units:

`to_metric_time(ts: int, src: TimeUnit, dst: TimeUnit)`

or something similar.

benjaminrwilson (changes requested):

argoverse/data_loading/synchronization_database.py
johnwlambert: typing_extensions.Final?

argoverse/data_loading/synchronization_database.py
johnwlambert:
```python
MAX_LIDAR_RING_CAM_TIMESTAMP_DIFF = to_metric_time(RING_CAMERA_SHUTTER_INTERVAL_MS / 2), src=Millisecond, dst=Nanosecond)
```

PR #219, opened 2021-05-07
Increase readability by placing global constants in a config

johnwlambert:
Currently, global constants like frame rate and image dimensions are scattered across the codebase, including in `data_loading/synchronization_database.py` and  `utils/camera_stats.py`. This PR places them in a more intuitive place for the user -- a single config file -- increasing readability for the codebase.

### DIFF
--- a/argoverse/config/argoverse1.1.yml
+++ b/argoverse/config/argoverse1.1.yml
@@ -1,0 +1,54 @@
+SensorDatasetConfig:
+  _target_: argoverse.sensor_dataset_config.SensorDatasetConfig
+  dataset_name: argoverse1.1
+  ring_cam_fps: 30
+  stereo_cam_fps: 5
+
+  # all ring cameras are `landscape` aspect ratio
+  sensors:
+    _target_: argoverse.sensor_dataset_config.SensorDimensions
+    
+    ring_front_center: 
+      _target_: argoverse.sensor_dataset_config.ImageDimensions
+      height: 1200
+      width: 1920
+
+    ring_front_left:
+      _target_: argoverse.sensor_dataset_config.ImageDimensions
+      height: 1200
+      width: 1920
+
+    ring_front_right:
+      _target_: argoverse.sensor_dataset_config.ImageDimensions
+      height: 1200
+      width: 1920
+
+    ring_side_left:
+      _target_: argoverse.sensor_dataset_config.ImageDimensions
+      height: 1200
+      width: 1920
+
+    ring_side_right:
+      _target_: argoverse.sensor_dataset_config.ImageDimensions
+      height: 1200
+      width: 1920
+
+    ring_rear_left:
+      _target_: argoverse.sensor_dataset_config.ImageDimensions
+      height: 1200
+      width: 1920
+
+    ring_rear_right:
+      _target_: argoverse.sensor_dataset_config.ImageDimensions
+      height: 1200
+      width: 1920
+
+    stereo_front_right:
+      _target_: argoverse.sensor_dataset_config.ImageDimensions
+      height: 2056
+      width: 2464
+
+    stereo_front_left:
+      _target_: argoverse.sensor_dataset_config.ImageDimensions
+      height: 2056
+      width: 2464
--- a/argoverse/sensor_dataset_config.py
+++ b/argoverse/sensor_dataset_config.py
@@ -1,0 +1,35 @@
+
+from typing import NamedTuple, Optional, Tuple
+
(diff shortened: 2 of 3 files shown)


argoverse/sensor_dataset_config.py
```diff
@@ -0,0 +1,41 @@
+from typing import NamedTuple, Optional
+
+from hydra.experimental import compose, initialize_config_module
+from hydra.utils import instantiate
+
+
+class ImageDimensions(NamedTuple):
```
benjaminrwilson, line 7: Can we make these `dataclasses` now that we support Python >3.6? I believe `omegaconf` uses `dataclasses` for their Structured Configs which `hydra` uses in the backend.

argoverse/config/argoverse1.1.yml
```diff
@@ -0,0 +1,54 @@
+SensorDatasetConfig:
+  _target_: argoverse.sensor_dataset_config.SensorDatasetConfig
+  dataset_name: argoverse1.1
```
benjaminrwilson, line 3: What do you think about having a separate version field? I think it might be cleaner.

argoverse/config/argoverse1.1.yml
```diff
@@ -0,0 +1,54 @@
...
+  ring_cam_fps: 30
+  stereo_cam_fps: 5
+
+  # all ring cameras are `landscape` aspect ratio
+  sensors:
+    _target_: argoverse.sensor_dataset_config.SensorDimensions
+    
+    ring_front_center: 
+      _target_: argoverse.sensor_dataset_config.ImageDimensions
```
benjaminrwilson, line 12: Hmm, I would consider encapsulating the functionality of `ImageDimensions` into an entire `SensorConfig` (or something similar) with additional functionality.

argoverse/sensor_dataset_config.py
```diff
@@ -0,0 +1,41 @@
...
+    """Dimensions are provided in pixels."""
+
+    height: int
+    width: int
+
+
+class SensorDimensions(NamedTuple):
```
benjaminrwilson, line 14: For configuration classes, what do you think about maintained the `Config` suffix -- e.g., `SensorDimensionsConfig`? I've seen that consistently used in other repositories related to Pytorch.

benjaminrwilson (changes requested):

argoverse/sensor_dataset_config.py
johnwlambert: I like the `Config` suffix -- should be incorporated now

argoverse/sensor_dataset_config.py
```diff
@@ -0,0 +1,41 @@
...
+class SensorDatasetConfig(NamedTuple):
+    """Global constants regarding frame rate and image dimensions."""
+
+    dataset_name: str
+    ring_cam_fps: int
+    stereo_cam_fps: int
+    sensors: SensorSuiteConfig
+
+
+DATASET_NAME = "argoverse1.1"
```
benjaminrwilson, line 37: I think "argoverse-v1.1" is a bit more readable. What are your thoughts?

johnwlambert:
Note: structured configs w/ defaults and type validation to be considered in follow-up PR

argoverse/sensor_dataset_config.py
```diff
@@ -0,0 +1,41 @@
...
+
+with hydra.initialize_config_module(config_module="argoverse.config"):
```
benjaminrwilson, line 39: What is this being used for?

argoverse/config/argoverse1.1.yaml
```diff
@@ -0,0 +1,54 @@
+SensorDatasetConfig:
```
benjaminrwilson, line 1: Do we need this to be nested?

argoverse/sensor_dataset_config.py
johnwlambert: agreed -- I'll update that

argoverse/sensor_dataset_config.py
johnwlambert: This is being instantiated here so we can do the following elsewhere:
```python
from argoverse.sensor_dataset_config import ArgoverseConfig
```
and we'll immediately have access to an instantiated class

argoverse/config/argoverse1.1.yaml
johnwlambert: I think the hierarchy makes it a bit more readable -- a class with 4 fields instead of 11

PR #223, opened 2021-05-12
Re-normalize quaternions if norm !=1, instead of throwing error

johnwlambert:
A user may provide a quaternion that does not have exactly unit norm. Instead of throwing an error (previous behavior), we can simply re-normalize the quaternion and gracefully proceed.

### DIFF
--- a/argoverse/utils/transform.py
+++ b/argoverse/utils/transform.py
@@ -7,8 +7,12 @@ while avoiding singularities or discontinuities (e.g. gimbal lock).
 We require that the quaternions are normalized beforehand to be unit-length.
 """
 
+import logging
+
 import numpy as np
 from scipy.spatial.transform import Rotation
+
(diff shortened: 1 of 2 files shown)


argoverse/utils/transform.py
```diff
@@ -25,7 +25,9 @@ def quat2rotmat(q: np.ndarray) -> np.ndarray:
     Returns:
         R: Array of shape (3, 3) representing a rotation matrix.
     """
-    assert np.isclose(np.linalg.norm(q), 1.0, atol=1e-12)
+    if not np.isclose(np.linalg.norm(q), 1.0, atol=1e-12):
```
benjaminrwilson, line 28: Should we log this so the user knows? Additionally, is it okay if we fail if they pass in all zeroes for coefficients?

johnwlambert: it's logged now

PR #225, opened 2021-05-20
Update MANIFEST.in

schmidt-ju:
Added argoverse-v1.1.yaml to MANIFEST.In as included. This fixes the installation of the argoverse-api via pip.

Without this fix, the yaml file is missing and the api cannot be used properly.

Best regards

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,5 @@
 # include the LICENSE
 include LICENSE
+
+# include argoverse-v1.1.yaml in pip installation
+include argoverse/config/argoverse-v1.1.yaml


johnwlambert:
Hi @schmidt-ju, thanks for the fix. We overlooked including this file. I've confirmed that this resolves the pip install so that 
```
pip install argoverse @ git+https://github.com/argoai/argoverse-api.git@master
```
will work again

@jagjeet-singh and @benjaminrwilson would you mind reviewing this and approving the workflow, so CI can run?

jagjeet-singh:
LGTM

PR #226, opened 2021-05-24
Limit max number of scores associated with max boxes.

benjaminrwilson:
Fixes two issues:

1. Corrects typing of args in `rank`.
2. The scores array was incorrect when the number of detections exceeded the maximum allowed.
3. Scores are sorted now in the case that someone submits unordered detections.

### DIFF
--- a/argoverse/evaluation/detection/utils.py
+++ b/argoverse/evaluation/detection/utils.py
@@ -157,8 +157,10 @@ def accumulate(
         logger.info(f"{dt_filtered.shape[0]} detections")
         logger.info(f"{gt_filtered.shape[0]} ground truth")
         if dt_filtered.shape[0] > 0:
-            ranked_detections, scores = rank(dt_filtered)
-            metrics = assign(ranked_detections, gt_filtered, cfg)
+            ranked_label_records = rank(dt_filtered)
+            metrics = assign(ranked_label_records, gt_filtered, cfg)
+
+            scores = [[record.score] for record in ranked_label_records.tolist()]
             cls_to_accum[class_name] = np.hstack((metrics, scores))
 
         cls_to_ninst[class_name] = gt_filtered.shape[0]
@@ -319,27 +321,24 @@ def filter_instances(
     return filtered_instances
 
 
-def rank(dts: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
-    """Get the rankings for the detections, according to detector confidence.
+def rank(label_records: np.ndarray) -> np.ndarray:
+    """Rank the `ObjectLabelRecord` objects in descending order by score.
 
     Args:
-        dts: Detections (N,).
+        label_records: Array of `ObjectLabelRecord` objects. (N,).
 
     Returns:
-        ranks: Ranking for the detections (N,).
-        scores: Detection scores (N,).
+        ranked_label_records: Array of `ObjectLabelRecord` objects ranked by score. (N,).
     """
-    scores = np.array([dt.score for dt in dts.tolist()])
+    scores = np.array([dt.score for dt in label_records.tolist()])
     ranks = scores.argsort()[::-1]
 
-    ranked_detections = dts[ranks]
-    ranked_scores = scores[ranks, None]
+    ranked_label_records = label_records[ranks]
 
     # Ensure the number of boxes considered per class is at most `MAX_NUM_BOXES`.
-    if ranked_detections.shape[0] > MAX_NUM_BOXES:
-        ranked_detections = ranked_detections[:MAX_NUM_BOXES]
-        ranked_scores = ranked_scores[:MAX_NUM_BOXES]
-    return ranked_detections, ranked_scores
+    if ranked_label_records.shape[0] > MAX_NUM_BOXES:
+        ranked_label_records = ranked_label_records[:MAX_NUM_BOXES]
+    return ranked_label_records
 
 
 def interp(prec: np.ndarray, method: InterpType = InterpType.ALL) -> np.ndarray:
--- a/argoverse/evaluation/detection/utils.py
+++ b/argoverse/evaluation/detection/utils.py
@@ -329,7 +329,7 @@ def rank(dts: np.ndarray) -> np.ndarray:
         dts: Array of `ObjectLabelRecord` objects. (N,).
 
     Returns:
-        ranked_dts: Array of `ObjectLabelRecord` objects ranked by score. (MAX_NUM_BOXES,).
+        ranked_dts: Array of `ObjectLabelRecord` objects ranked by score. (N,) where N <= MAX_NUM_BOXES
     """
     scores = np.array([dt.score for dt in dts.tolist()])
     ranks = scores.argsort()[::-1]
--- a/argoverse/evaluation/detection/utils.py
+++ b/argoverse/evaluation/detection/utils.py
@@ -92,7 +92,7 @@ class DetectionCfg(NamedTuple):
     dt_classes: List[str] = COMPETITION_CLASSES
     dt_metric: FilterMetric = FilterMetric.EUCLIDEAN
     max_dt_range: float = 100.0  # Meters
-    save_figs: bool = True
+    save_figs: bool = False
     tp_normalization_terms: np.ndarray = np.array([tp_thresh, MAX_SCALE_ERROR, MAX_YAW_ERROR])
     summary_default_vals: np.ndarray = np.array([MIN_AP, tp_thresh, MAX_NORMALIZED_ASE, MAX_NORMALIZED_AOE, MIN_CDS])
     eval_only_roi_instances: bool = True
--- a/argoverse/evaluation/detection/utils.py
+++ b/argoverse/evaluation/detection/utils.py
@@ -92,10 +92,10 @@ class DetectionCfg(NamedTuple):
     dt_classes: List[str] = COMPETITION_CLASSES
     dt_metric: FilterMetric = FilterMetric.EUCLIDEAN
     max_dt_range: float = 100.0  # Meters
-    save_figs: bool = False
+    save_figs: bool = True
     tp_normalization_terms: np.ndarray = np.array([tp_thresh, MAX_SCALE_ERROR, MAX_YAW_ERROR])
     summary_default_vals: np.ndarray = np.array([MIN_AP, tp_thresh, MAX_NORMALIZED_ASE, MAX_NORMALIZED_AOE, MIN_CDS])
-    eval_only_roi_instances: bool = True
+    eval_only_roi_instances: bool = False
     map_root: _PathLike = Path(__file__).parent.parent.parent.parent / "map_files"  # argoverse-api/map_files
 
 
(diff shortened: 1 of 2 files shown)


argoverse/evaluation/detection/utils.py
```diff
@@ -333,10 +329,17 @@ def rank(dts: List[ObjectLabelRecord]) -> Tuple[np.ndarray, np.ndarray]:
         ranks: Ranking for the detections (N,).
         scores: Detection scores (N,).
     """
-    scores = np.array([dt.score for dt in dts])
+    scores = np.array([dt.score for dt in dts.tolist()])
```
benjaminrwilson, line 332: This was changed to satisfy `mypy`.

argoverse/evaluation/detection/utils.py
```diff
@@ -323,7 +319,7 @@ def filter_instances(
     return filtered_instances
 
 
-def rank(dts: List[ObjectLabelRecord]) -> Tuple[np.ndarray, np.ndarray]:
+def rank(dts: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
```
johnwlambert, line 322: Hi @benjaminrwilson , since the function signature is a bit less informative now (albeit more correct), could you expand the `dts` description to:
```python
    Args:
        dts: Object array of shape (N,) representing detections. Each object is an `ObjectLabelRecord`
```

argoverse/evaluation/detection/utils.py
```diff
@@ -333,10 +329,17 @@ def rank(dts: List[ObjectLabelRecord]) -> Tuple[np.ndarray, np.ndarray]:
         ranks: Ranking for the detections (N,).
         scores: Detection scores (N,).
     """
-    scores = np.array([dt.score for dt in dts])
+    scores = np.array([dt.score for dt in dts.tolist()])
     ranks = scores.argsort()[::-1]
+
     ranked_detections = dts[ranks]
-    return ranked_detections, scores[:, np.newaxis]
+    ranked_scores = scores[ranks, None]
```
johnwlambert, line 336: I think `scores[ranks, None]` will not agree with the docstring, which says `(N,)` for scores?

johnwlambert: won't it be (N,1)?

johnwlambert: maybe we need to update the names of the return variables, in the docstring as well?

argoverse/evaluation/detection/utils.py
```diff
@@ -319,27 +321,24 @@ def filter_instances(
...
 
 
-def rank(dts: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
-    """Get the rankings for the detections, according to detector confidence.
+def rank(label_records: np.ndarray) -> np.ndarray:
+    """Rank the `ObjectLabelRecord` objects in descending order by score.
 
     Args:
-        dts: Detections (N,).
+        label_records: Array of `ObjectLabelRecord` objects. (N,).
```
johnwlambert, line 328: thanks for the updates, @benjaminrwilson. I liked the old name `dts`, more, though. Since it conveys the semantics. `label_records` sounds like GT labels, which they are not

argoverse/evaluation/detection/utils.py
```diff
@@ -157,8 +157,10 @@ def accumulate(
         logger.info(f"{dt_filtered.shape[0]} detections")
         logger.info(f"{gt_filtered.shape[0]} ground truth")
         if dt_filtered.shape[0] > 0:
-            ranked_detections, scores = rank(dt_filtered)
-            metrics = assign(ranked_detections, gt_filtered, cfg)
+            ranked_label_records = rank(dt_filtered)
```
johnwlambert, line 160: prefer `ranked_detections` more here, like it was before

argoverse/evaluation/detection/utils.py
```diff
@@ -212,10 +214,6 @@ def assign(dts: np.ndarray, gts: np.ndarray, cfg: DetectionCfg) -> np.ndarray:
             of true positive thresholds used for AP computation and S is the number of true positive errors.
     """
 
-    # Ensure the number of boxes considered per class is at most `MAX_NUM_BOXES`.
```
tagarwal-argoai, line 215: Are we removing this intentionally and does the downstream code handle this change well?

argoverse/evaluation/detection/utils.py
```diff
@@ -157,8 +157,10 @@ def accumulate(
         logger.info(f"{dt_filtered.shape[0]} detections")
         logger.info(f"{gt_filtered.shape[0]} ground truth")
         if dt_filtered.shape[0] > 0:
-            ranked_detections, scores = rank(dt_filtered)
-            metrics = assign(ranked_detections, gt_filtered, cfg)
+            ranked_dts = rank(dt_filtered)
+            metrics = assign(ranked_dts, gt_filtered, cfg)
+
+            scores = [[record.score] for record in ranked_dts.tolist()]
```
johnwlambert, line 163: seems like this extra for-loop is less efficient than just returning the scores earlier, right?

Might be nice to avoid it

johnwlambert: Maybe we can still return the scores?

argoverse/evaluation/detection/utils.py
```diff
@@ -323,20 +321,24 @@ def filter_instances(
     return filtered_instances
 
 
-def rank(dts: List[ObjectLabelRecord]) -> Tuple[np.ndarray, np.ndarray]:
-    """Get the rankings for the detections, according to detector confidence.
+def rank(dts: np.ndarray) -> np.ndarray:
+    """Rank the `ObjectLabelRecord` objects in descending order by score.
```
johnwlambert, line 325: This seems like an implementation detail to me. I would prefer:
```python
"""Rank the detections in descending order according to score (detector confidence).
```

argoverse/evaluation/detection/utils.py
tagarwal-argoai: Nvm, you have handled it below.

argoverse/evaluation/detection/utils.py
```diff
@@ -323,20 +321,24 @@ def filter_instances(
...
 
     Args:
-        dts: Detections (N,).
+        dts: Array of `ObjectLabelRecord` objects. (N,).
 
     Returns:
-        ranks: Ranking for the detections (N,).
-        scores: Detection scores (N,).
+        ranked_dts: Array of `ObjectLabelRecord` objects ranked by score. (N,).
```
tagarwal-argoai, line 331: I think the shape of the numpy should be `(MAX_NUM_BOXES, )`

tagarwal-argoai (approved):
Looks good to me!

argoverse/evaluation/detection/utils.py
johnwlambert: N could be << MAX_NUM_BOXES, but it would be clarify in the docstring, maybe something like:
```
ranked_dts: Array of `ObjectLabelRecord` objects ranked by score. (N,) where N <= MAX_NUM_BOXES
```

argoverse/evaluation/detection/utils.py
benjaminrwilson: Are you okay with the sizing mismatch where `ranked_dts` is `(N,)` and `scores` is `(N, 1)`?

johnwlambert: Is there a reason we need (N,1) here? If so, how about we return (N,) for both, and just reshape to (N,1) later where it's used?

argoverse/evaluation/detection/utils.py
benjaminrwilson: Thanks for pointing this out, guys.

argoverse/evaluation/detection/utils.py
benjaminrwilson: Resolved.

johnwlambert (approved):
LGTM

argoverse/evaluation/detection/utils.py
```diff
@@ -323,20 +320,28 @@ def filter_instances(
...
+    scores = np.array([dt.score for dt in dts.tolist()])
     ranks = scores.argsort()[::-1]
-    ranked_detections = dts[ranks]
-    return ranked_detections, scores[:, np.newaxis]
+
+    ranked_dts = dts[ranks]
+    ranked_scores = scores[ranks]
+
+    # Ensure the number of boxes considered per class is at most `MAX_NUM_BOXES`.
+    if ranked_dts.shape[0] > MAX_NUM_BOXES:
```
jagjeet-singh, line 341: [nit] We can avoid the if condition here.
Just `return ranked_dts[:MAX_NUM_BOXES], ranked_scores[:MAX_NUM_BOXES]`

argoverse/evaluation/detection/utils.py
```diff
@@ -157,9 +157,10 @@ def accumulate(
         logger.info(f"{dt_filtered.shape[0]} detections")
         logger.info(f"{gt_filtered.shape[0]} ground truth")
         if dt_filtered.shape[0] > 0:
-            ranked_detections, scores = rank(dt_filtered)
-            metrics = assign(ranked_detections, gt_filtered, cfg)
-            cls_to_accum[class_name] = np.hstack((metrics, scores))
+            ranked_dts, ranked_scores = rank(dt_filtered)
```
jagjeet-singh, line 160: [nit] `detections` is more readable than `dts`. Looks like we changed to latter in multiple places.

argoverse/evaluation/detection/utils.py
```diff
@@ -323,20 +320,28 @@ def filter_instances(
     return filtered_instances
 
 
-def rank(dts: List[ObjectLabelRecord]) -> Tuple[np.ndarray, np.ndarray]:
-    """Get the rankings for the detections, according to detector confidence.
+def rank(dts: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+    """Rank the detections in descending order according to score (detector confidence).
```
jagjeet-singh, line 324: Should we add a test case for this in the unit tests?

johnwlambert: indeed we should

argoverse/evaluation/detection/utils.py
johnwlambert: good point

argoverse/evaluation/detection/utils.py
benjaminrwilson: If we do go this route, I think we should switch all of the variables over to the more verbose naming (e.g., `dt_filtered` -> `detections_filtered`). What do you think?

argoverse/evaluation/detection/utils.py
benjaminrwilson: Agreed. I'll add one.

argoverse/evaluation/detection/utils.py
johnwlambert: how about we keep it generally as it was before? I think `dt_filtered` is actually simpler to read than `detections_filtered`. That variable name seems to be getting too long

argoverse/evaluation/detection/utils.py
benjaminrwilson: I think the former is a tiny bit faster from benchmarking, but they're both on the order of `1e-7 - 1e-6`, so I personally think removing the conditional would look better.

PR #228, opened 2021-05-24
Add pyproject.toml

benjaminrwilson:
This PR adds a `pyproject.toml` file to consolidate styling tools such as `isort` and `black`.

- `pyproject.toml` was introduced in [PEP 518](https://www.python.org/dev/peps/pep-0518/) for build system requirements.
- Allows for a unified configuration file for a host of tools, e.g. `isort`, `black`, etc.

Each "block" within this file specifies a different type of options. For example, `isort` is written as:

```
[tool.isort]

known_first_party = "argoverse"
line_length = 120
multi_line_output = 3
profile = "black"
```

The options above should have a 1-to-1 correspondence with `isort.cfg`.

### DIFF
--- a/argoverse/evaluation/detection/utils.py
+++ b/argoverse/evaluation/detection/utils.py
@@ -338,10 +338,7 @@ def rank(dts: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
     ranked_scores = scores[ranks]
 
     # Ensure the number of boxes considered per class is at most `MAX_NUM_BOXES`.
-    if ranked_dts.shape[0] > MAX_NUM_BOXES:
-        ranked_dts = ranked_dts[:MAX_NUM_BOXES]
-        ranked_scores = ranked_scores[:MAX_NUM_BOXES]
-    return ranked_dts, ranked_scores
+    return ranked_dts[:MAX_NUM_BOXES], ranked_scores[:MAX_NUM_BOXES]
 
 
 def interp(prec: np.ndarray, method: InterpType = InterpType.ALL) -> np.ndarray:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,5 @@ line_length = 120
 multi_line_output = 3
 profile = "black"
 
-
 [tool.black]
 line-length = 120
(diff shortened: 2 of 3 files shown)


.isort.cfg
```diff
@@ -1,5 +0,0 @@
-[settings]
```
tagarwal-argoai, line 1: Can you explain what the purpose of the file was before this change and why it has been removed?

tagarwal-argoai, line 1: Can you explain what the purpose of the file was before this change and why it has been removed?

argoverse/evaluation/detection/utils.py
```diff
@@ -338,10 +338,7 @@ def rank(dts: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
     ranked_scores = scores[ranks]
 
     # Ensure the number of boxes considered per class is at most `MAX_NUM_BOXES`.
-    if ranked_dts.shape[0] > MAX_NUM_BOXES:
```
tagarwal-argoai, line 341: Can you change the base of this PR to point to the branch used in 227 as I am sure this change is not part of this PR?

pyproject.toml
```diff
@@ -0,0 +1,8 @@
+[tool.isort]
```
tagarwal-argoai, line 1: Can you add a description to the PR on how to parse this file and what the different fields in this file indicate?

tagarwal-argoai (commented):
Left a couple of clarifying questions.

pyproject.toml
benjaminrwilson: Added some info. We should double check to see if we need `known_third_party`. Maintaining this list seems like it could become a lot of overhead.

tests/test_eval_detection.py
```diff
@@ -215,8 +216,26 @@ def test_accumulate() -> None:
         cls_to_accum["VEHICLE"]
         == np.array(
             [
-                [1.0, 1.0, 1.0, 1.0, expected_ATE, expected_ASE, expected_AOE, expected_AP],
-                [1.0, 1.0, 1.0, 1.0, expected_ATE, expected_ASE, expected_AOE, expected_AP],
+                [
```
johnwlambert, line 219: can we remove these changes from this PR, Ben?

johnwlambert:
Ben, what are the advantages of pyproject.toml over our current repo setup?

PR #229, opened 2021-05-25
Optimize SE(3) transform_point_cloud() method

johnwlambert:
Currently our implementation is not very efficient. We allocate an array of ones to create homogenous points before the matrix multiply with the transformation matrix, which is unnecessary. The new implementation simply does `points @ R.T + t` which is all that is needed.

This is also how other libraries like GTSAM implement this (see [here](https://github.com/borglab/gtsam/blob/develop/gtsam/geometry/Pose3.cpp#L310)).

### DIFF
--- a/argoverse/utils/se3.py
+++ b/argoverse/utils/se3.py
@@ -37,9 +37,7 @@ class SE3:
 
     def inverse_transform_point_cloud(self, point_cloud: np.ndarray) -> np.ndarray:
         """Undo the translation and then the rotation (Inverse SE(3) transformation)."""
-        point_cloud = point_cloud.copy()
-        point_cloud -= self.translation
-        return point_cloud @ self.rotation
+        return (point_cloud.copy() - self.translation) @ self.rotation
 
     def inverse(self) -> "SE3":
         """Return the inverse of the current SE3 transformation.
--- a/argoverse/utils/se3.py
+++ b/argoverse/utils/se3.py
@@ -27,16 +27,13 @@ class SE3:
         """Apply the SE(3) transformation to this point cloud.
 
         Args:
-            point_cloud: Array of shape (N, 3)
+            point_cloud: Array of shape (N, 3). If the transform represents dst_SE3_src,
+                then point_cloud should consist of points in frame `src`
 
         Returns:
-            transformed_point_cloud: Array of shape (N, 3)
+            Array of shape (N, 3) representing the transformed point cloud, i.e. points in frame `dst`
         """
-        # convert to homogeneous
-        num_pts = point_cloud.shape[0]
-        homogeneous_pts = np.hstack([point_cloud, np.ones((num_pts, 1))])
-        transformed_point_cloud = homogeneous_pts.dot(self.transform_matrix.T)
-        return transformed_point_cloud[:, :3]
+        return point_cloud @ self.rotation.T + self.translation
 
     def inverse_transform_point_cloud(self, point_cloud: np.ndarray) -> np.ndarray:
         """Undo the translation and then the rotation (Inverse SE(3) transformation)."""
--- a/argoverse/utils/se3.py
+++ b/argoverse/utils/se3.py
@@ -39,7 +39,7 @@ class SE3:
         """Undo the translation and then the rotation (Inverse SE(3) transformation)."""
         point_cloud = point_cloud.copy()
         point_cloud -= self.translation
-        return point_cloud.dot(self.rotation)
+        return point_cloud @ self.rotation
 
     def inverse(self) -> "SE3":
         """Return the inverse of the current SE3 transformation.
--- a/tests/test_se3.py
+++ b/tests/test_se3.py
@@ -1,6 +1,7 @@
 # <Copyright 2019, Argo AI, LLC. Released under the MIT license.>
 """Tests to verify the SE3 class works correctly."""
 
+import time
 from typing import Union
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


johnwlambert:
@benjaminrwilson could you take a look at this when you get a chance?

tests/test_se3.py
```diff
@@ -159,3 +160,45 @@ def test_SE3_chaining_transforms() -> None:
...
+    # x, y, z of cuboid center
+    t = np.array([-34.7128603513203, 5.29461762417753, 0.10328996181488])
+
+    # quaternion has order (w,x,y,z)
+    q = np.array([0.700322174885275, 0.0, 0.0, -0.713826905743933])
+    R = quat2rotmat(q)
+
+    dst_se3_src = SE3(rotation=R.copy(), translation=t.copy())
+
+    start = time.time()
```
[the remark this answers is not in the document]
benjaminrwilson: This would probably be a better solution, though: https://pypi.org/project/pytest-benchmark/

argoverse/utils/se3.py
```diff
@@ -27,16 +27,13 @@ def transform_point_cloud(self, point_cloud: np.ndarray) -> np.ndarray:
...
         """
-        # convert to homogeneous
-        num_pts = point_cloud.shape[0]
-        homogeneous_pts = np.hstack([point_cloud, np.ones((num_pts, 1))])
-        transformed_point_cloud = homogeneous_pts.dot(self.transform_matrix.T)
-        return transformed_point_cloud[:, :3]
+        return point_cloud @ self.rotation.T + self.translation
 
     def inverse_transform_point_cloud(self, point_cloud: np.ndarray) -> np.ndarray:
         """Undo the translation and then the rotation (Inverse SE(3) transformation)."""
```
tagarwal-argoai, line 39: Could we also edit this function accordingly so that we can do the operation in one line.

tests/test_se3.py
```diff
@@ -159,3 +160,45 @@ def test_SE3_chaining_transforms() -> None:
...
+    def transform_point_cloud_slow(point_cloud: np.ndarray, transform_matrix: np.ndarray) -> np.ndarray:
+        """Slow, unoptimized method -- allocated unnecessary array for homogeneous coordinates."""
+        # convert to homogeneous
+        num_pts = point_cloud.shape[0]
+        homogeneous_pts = np.hstack([point_cloud, np.ones((num_pts, 1))])
+        transformed_point_cloud = homogeneous_pts.dot(transform_matrix.T)
+        return transformed_point_cloud[:, :3]
+
+    num_trials = 1000
+    num_pts = 100000
```
tagarwal-argoai, line 180: Can we move this up as module level constants?

tests/test_se3.py
```diff
@@ -159,3 +160,45 @@ def test_SE3_chaining_transforms() -> None:
...
+    pts_src = np.random.randn(num_pts, 3)
+
+    # x, y, z of cuboid center
+    t = np.array([-34.7128603513203, 5.29461762417753, 0.10328996181488])
+
+    # quaternion has order (w,x,y,z)
+    q = np.array([0.700322174885275, 0.0, 0.0, -0.713826905743933])
+    R = quat2rotmat(q)
```
tagarwal-argoai, line 188: The docstring for this function mentions that SE(3) transformations are chosen arbitrarily which is not the case here. Can you confirm? 

Also how do you feel about this moving this outside the test function and maybe take it as an argument?

tests/test_se3.py
```diff
@@ -159,3 +160,45 @@ def test_SE3_chaining_transforms() -> None:
...
+    end = time.time()
+    optimized_duration = end - start
+
+    start = time.time()
+    for i in range(num_trials):
+        pts_dst = transform_point_cloud_slow(pts_src, dst_se3_src.transform_matrix)
+    end = time.time()
+    unoptimized_duration = end - start
+
+    assert optimized_duration < unoptimized_duration
```
tagarwal-argoai, line 204: Nice!

tagarwal-argoai (approved):
Looks good overall, but I will let @benjaminrwilson shine in for the final review.

benjaminrwilson:
@johnwlambert , what are you thinking in relation to the above comments?

tests/test_se3.py
johnwlambert: So the SE(3) transformation here is actually just an arbitrary one, so I think the docstring is correct.

argoverse/utils/se3.py
johnwlambert: Thanks for the review Tanmay -- I'm not sure this would be more readable on one line, since it would look like
```
    def inverse_transform_point_cloud(self, point_cloud: np.ndarray) -> np.ndarray:
        """Undo the translation and then the rotation (Inverse SE(3) transformation)."""
        return (point_cloud.copy() - self.translation) @ self.rotation
```
@benjaminrwilson what do you think?

tests/test_se3.py
johnwlambert: not sure if they apply to the module, since they apply to just a single test out of 8. I'm worried that it might confuse people, since they might assume it applies to each unit test

argoverse/utils/se3.py
benjaminrwilson: I prefer the one-liner because it makes it easier for me to read it as an equation. On a separate note, do you think we should have an option to transform `point_cloud` in-place? What do you think is the most intuitive design decision when comparing expected behavior to speed?

benjaminrwilson (approved):

PR #232, opened 2021-05-27
Added poetry + conda install.

benjaminrwilson:
This PR addresses a few different things, but will likely require a bit of additional discussion + additions.

1. Workflow support for Conda builds.
2. Migrated from `tox` to `nox`.
3. Added a few default configuration files for CI/CD.
4. Fixed various style errors that showed up in other files.


pyproject.toml
```diff
@@ -0,0 +1,38 @@
+[tool.black]
+line-length = 120
+
+[tool.poetry]
+name = "argoverse-api"
+version = "0.1.0"
+description = ""
+authors = ["Your Name <you@example.com>"]
```
benjaminrwilson, line 8: We need to fill in something here.

noxfile.py
```diff
@@ -0,0 +1,142 @@
...
+            "--dev",
+            "--without-hashes",
+            "--format=requirements.txt",
+            f"--output={requirements.name}",
+            external=True,
+        )
+        session.install(f"--constraint={requirements.name}", *args, **kwargs)
+
+
+@nox.session(python=["3.7", "3.8"])
```
benjaminrwilson, line 43: I guess we could run against `3.9` as well.

PR #235, opened 2021-06-04
Hydra fix.

benjaminrwilson:
There is currently an issue when initializing `hydra` from another library when using `argoverse`. This will mitigate this issue until another solution can be determined (if needed).

### DIFF
--- a/argoverse/sensor_dataset_config.py
+++ b/argoverse/sensor_dataset_config.py
@@ -1,8 +1,6 @@
 from dataclasses import dataclass
 from typing import Optional
 
-import hydra
-from hydra.utils import instantiate
 from omegaconf import MISSING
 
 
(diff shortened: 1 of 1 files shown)


benjaminrwilson:
@johnwlambert -- whenever you get a chance.

argoverse/sensor_dataset_config.py
```diff
@@ -49,8 +47,27 @@ class SensorDatasetConfig:
...
+    STEREO_CAMERA_WIDTH: int = 2464
+
+    dataset_name: str = "argoverse"
+    ring_cam_fps: int = 30
+    stereo_cam_fps: int = 5
 
-with hydra.initialize_config_module(config_module="argoverse.config"):
-    cfg = hydra.compose(config_name=f"{DATASET_NAME}.yaml")
-    ArgoverseConfig: SensorDatasetConfig = instantiate(cfg.SensorDatasetConfig)
+    camera_sensors: SensorSuiteConfig = SensorSuiteConfig(
```
benjaminrwilson, line 62: Repeated code is frustrating here ... but can't use `__post_init__` since the dataclass is frozen.

tagarwal-argoai: I am missing something.. What's being repeated here? And can we not unfreeze the dataclass?

tagarwal-argoai (commented):
Left a small comment.

argoverse/sensor_dataset_config.py
benjaminrwilson: Well, I just mean that we could generate this more succinctly in a loop. Maybe not an issue. We could unfreeze, but it's supposed to be immutable since it's a config.

tagarwal-argoai: I think this is fine and not a major thing. Let's leave this for now.

tagarwal-argoai (approved):
LGTM!

argoverse/sensor_dataset_config.py
```diff
@@ -49,8 +47,27 @@ class SensorDatasetConfig:
     camera_sensors: SensorSuiteConfig = MISSING
 
 
-DATASET_NAME = "argoverse-v1.1"
```
johnwlambert, line 52: could we put this in the docstring, instead of removing it completely?

benjaminrwilson: Could we move it to the `dataset_name` attribute on line 58?

johnwlambert: so i didnt quite follow why we need the dataset name on line 58

argoverse/sensor_dataset_config.py
```diff
@@ -49,8 +47,27 @@ class SensorDatasetConfig:
...
+@dataclass(frozen=True)
+class ArgoverseConfig(SensorDatasetConfig):
+    RING_CAMERA_HEIGHT: int = 1200
+    RING_CAMERA_WIDTH: int = 1920
+
+    STEREO_CAMERA_HEIGHT: int = 2056
+    STEREO_CAMERA_WIDTH: int = 2464
+
+    dataset_name: str = "argoverse"
```
johnwlambert, line 58: i don't think this is specific enough to be helpful. if it was "argoverse-tracking" or "argoverse-v1.1", perhaps.

johnwlambert: i vote we keep it as before
```
  dataset_name: argoverse-v1.1
```

benjaminrwilson: 👍

johnwlambert (approved):

johnwlambert:
Do you mind pinging Omry in the thread here about the issue you saw, https://github.com/facebookresearch/hydra/issues/1628?

PR #238, opened 2021-06-14
Throw exception if input to Sim(2)'s `transform_from()` is not 2d

johnwlambert:
Input must be (N,2) not (2,) in shape.

Using the previous code, a (2,) array vs. (1,2) array would yield different results:
```python
import numpy as np

from argoverse.utils.sim2 import Sim2


def rotmat2d(theta_deg: float) -> np.ndarray:
	""" """
	theta_rad = np.deg2rad(theta_deg)
	c = np.cos(theta_rad)
	s = np.sin(theta_rad)

	R = np.array([[c,-s],[s,c]])

	return R

R = rotmat2d(-13)
t = np.array([-5,1])
s = 1.5
aSb = Sim2(R, t, s)

pt_b = np.array([1,0])
pt_a = aSb.transform_from(pt_b)
print(pt_a)

pt_b = np.array([1,0]).reshape(1,2)
pt_a = aSb.transform_from(pt_b)
print(pt_a)
```
used to print:
```
[[-6.03844491  2.96155509]
 [-7.83742659  1.16257341]]
[[-6.03844491  1.16257341]]
```
Instead, there should be the same (1,2) array as output.

Now, fixed via broadcasting.

### DIFF
--- a/argoverse/utils/sim2.py
+++ b/argoverse/utils/sim2.py
@@ -98,6 +98,8 @@ class Sim2:
         Returns:
             transformed_point_cloud: Nx2 array representing 2d points in frame B
         """
+        if not point_cloud.ndim == 2:
+            raise ValueError("Input point cloud is not 2-dimensional.")
         assert_np_array_shape(point_cloud, (None, 2))
         # (2,2) x (2,N) + (2,1) = (2,N) -> transpose
         transformed_point_cloud = (self.R_ @ point_cloud.T + self.t_.reshape(2, 1)).T
--- a/argoverse/utils/sim2.py
+++ b/argoverse/utils/sim2.py
@@ -102,7 +102,7 @@ class Sim2:
             raise ValueError("Input point cloud is not 2-dimensional.")
         assert_np_array_shape(point_cloud, (None, 2))
         # (2,2) x (2,N) + (2,1) = (2,N) -> transpose
-        transformed_point_cloud = (self.R_ @ point_cloud.T + self.t_.reshape(2, 1)).T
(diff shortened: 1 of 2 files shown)


argoverse/utils/sim2.py
```diff
@@ -98,6 +98,8 @@ def transform_from(self, point_cloud: np.ndarray) -> np.ndarray:
         Returns:
             transformed_point_cloud: Nx2 array representing 2d points in frame B
         """
+        if not point_cloud.ndim == 2:
```
benjaminrwilson, line 101: What is the motivation for overriding the default broadcasting behavior in `numpy`?

johnwlambert: I guess the main reason here is that a (2,) vector and a (1,2) vector will have very different results, and only the (1,2) case will be correct, in the way we've defined it.

johnwlambert: if we have a (2,2) array, we also have to be careful to make sure which direction we broadcast the sum with the translation over

johnwlambert:
Hi @tagarwal-argoai, would you mind taking a brief look at this? Thanks!

johnwlambert:
@benjaminrwilson I made a change so that the broadcasting works now.

But the docstring says the input has to be (N,2) to get (N,2) back out. So I think the ndim check is still important.

argoverse/utils/sim2.py
benjaminrwilson: Thanks for working on this, @johnwlambert. I was wondering what you thought about the following:

```
point_cloud = np.atleast_2d(point_cloud)
```

and then we could simplify by removing lines 101 - 103. Thoughts?

PR #240, opened 2021-06-23
Updated: Add factory functions for Sim(2) clasmethod and calibration helpers in loader

johnwlambert:
Updated variant of https://github.com/argoai/argoverse-api/pull/212, which is now closed

No longer have write access to the branch in https://github.com/argoai/argoverse-api/pull/212


argoverse/data_loading/simple_track_dataloader.py
```diff
@@ -26,7 +27,7 @@ def __init__(self, data_dir: str, labels_dir: str) -> None:
         self.sdb = SynchronizationDB(data_dir)
 
     def get_city_name(self, log_id: str) -> str:
-        """
+        """Return the name of the city where the log of interest was cpatured.
```
benjaminrwilson, line 30: nit: cpatured -> captured.

johnwlambert: good catch, should be fixed now

argoverse/utils/sim2.py
```diff
@@ -141,3 +144,38 @@ def transform_from(self, point_cloud: np.ndarray) -> np.ndarray:
     def transform_point_cloud(self, point_cloud: np.ndarray) -> np.ndarray:
         """Alias for `transform_from()`, for synchrony w/ API provided by SE(2) and SE(3) classes."""
         return self.transform_from(point_cloud)
+
+    def save_as_json(self, save_fpath: Union[str, "os.PathLike[str]"]) -> None:
```
benjaminrwilson, line 148: How do you want to handle the typing of `Union[str, "os.PathLike[str]"]`? In some files we're doing this: https://github.com/argoai/argoverse-api/blob/9ad7251e863590603c3f6064a45830dabd3a15e2/argoverse/utils/pkl_utils.py#L8

benjaminrwilson (commented):
Looks really good, @johnwlambert. I had one small comment related to how we handle typing of path-like objects.

argoverse/utils/sim2.py
johnwlambert: Thanks for the review @benjaminrwilson . I like the ` _PathLike = Union[str, "os.PathLike[str]"] ` macro. If you like it as well, I'll use it here too

benjaminrwilson: That sounds good. Is it recommended for us to repeat that macro in each file it's used?

PR #243, opened 2021-06-23
Fix matmul bug and add __repr__ method to Sim2

johnwlambert:
Previously, the `compose()` method had a bug ( R1 * R2 was used instead of the correct R1 @ R2) that was not caught since the unit test used identity matrices, for which the operations are identical.

This PR also adds a __repr__ method for helpful printing/debugging of the class representation.

### DIFF
--- a/argoverse/utils/sim2.py
+++ b/argoverse/utils/sim2.py
@@ -37,11 +37,16 @@ class Sim2:
 
     @property
     def theta_deg(self) -> float:
-        """Recover the rotation angle `theta` from the 2d rotation matrix.
+        """Recover the rotation angle `theta` (in degrees) from the 2d rotation matrix.
 
-        Note: tan(theta) = s/c = (opp/hyp) / (adj/hyp) = opp/adj
+        Note: the first column of the rotation matrix R provides sine and cosine of theta,
+            since R is encoded as [c,-s]
+                                  [s, c]
+
+        We use the following identity: tan(theta) = s/c = (opp/hyp) / (adj/hyp) = opp/adj
         """
-        theta_rad = np.arctan2(self.R_[1, 0], self.R_[0, 0])
+        c, s = self.R_[0, 0], self.R_[1, 0]
+        theta_rad = np.arctan2(s, c)
         return np.rad2deg(theta_rad)
 
     def __repr__(self) -> str:
--- a/argoverse/utils/sim2.py
+++ b/argoverse/utils/sim2.py
@@ -94,8 +94,21 @@ class Sim2:
         return T
 
     def compose(self, S: "Sim2") -> "Sim2":
-        """Composition with another Sim2."""
-        return Sim2(self.R_ @ S.R_, ((1.0 / S.s_) * self.t_) + self.R_ @ S.t_, self.s_ * S.s_)
+        """Composition with another Sim2.
+
+        This can be understood via block matrix multiplication, if self is parameterized as (R1,t1,s1)
+        and if `S` is parameterized as (R2,t2,s2):
+
+        [R1  t1]   [R2  t2]   [R1 @ R2   R1@t2 + t1/s2]
+        [0 1/s1] @ [0 1/s2] = [ 0          1/(s1*s2)  ]
+        """
+        # fmt: off
+        return Sim2(
+            R=self.R_ @ S.R_,
+            t=self.R_ @ S.t_ + ((1.0 / S.s_) * self.t_),
+            s=self.s_ * S.s_
+        )
+        # fmt: on
 
     def inverse(self) -> "Sim2":
         """Return the inverse."""
--- a/argoverse/utils/sim2.py
+++ b/argoverse/utils/sim2.py
@@ -54,7 +54,7 @@ class Sim2:
         return f"Angle (deg.): {self.theta_deg}, Trans.: {np.round(self.t_,2)}, Scale: {self.s_}"
 
     def __eq__(self, other: object) -> bool:
-        """Check for equality with other Sim(2) object"""
+        """Check for equality with other Sim(2) object."""
         if not isinstance(other, Sim2):
             return False
 
--- a/argoverse/utils/sim2.py
+++ b/argoverse/utils/sim2.py
@@ -47,7 +47,7 @@ class Sim2:
         """
         c, s = self.R_[0, 0], self.R_[1, 0]
         theta_rad = np.arctan2(s, c)
-        return np.rad2deg(theta_rad)
+        return float(np.rad2deg(theta_rad))
 
     def __repr__(self) -> str:
         """Return a human-readable string representation of the class."""
--- a/argoverse/utils/sim2.py
+++ b/argoverse/utils/sim2.py
@@ -45,7 +45,7 @@ class Sim2:
         return np.rad2deg(theta_rad)
 
     def __repr__(self) -> str:
-        """ """
+        """Return a human-readable string representation of the class."""
         return f"Angle (deg.): {self.theta_deg}, Trans.: {np.round(self.t_,2)}, Scale: {self.s_}"
 
     def __eq__(self, other: object) -> bool:
--- a/argoverse/utils/sim2.py
+++ b/argoverse/utils/sim2.py
@@ -71,12 +71,12 @@ class Sim2:
 
     @property
     def rotation(self) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


tests/test_sim2.py
```diff
@@ -195,6 +225,34 @@ def test_cannot_set_zero_scale() -> None:
         Sim2(R, t, s)
 
 
+def test_sim2_theta_deg_1() -> None:
+    """Ensure we can recover the rotation angle theta"""
```
benjaminrwilson, line 229: Nit: period and end of sentence for consistency.

tests/test_sim2.py
```diff
@@ -195,6 +225,34 @@ def test_cannot_set_zero_scale() -> None:
...
+    R = np.eye(2)
+    t = np.arange(2)
+    s = 10.5
+    aSb = Sim2(R, t, s)
+    assert aSb.theta_deg == 0
+
+
+def test_sim2_theta_deg_2() -> None:
+    """Ensure we can recover the rotation angle theta"""
```
benjaminrwilson, line 238: Nit: period and end of sentence for consistency.

argoverse/utils/sim2.py
```diff
@@ -77,7 +95,7 @@ def matrix(self) -> np.ndarray:
 
     def compose(self, S: "Sim2") -> "Sim2":
         """Composition with another Sim2."""
-        return Sim2(self.R_ * S.R_, ((1.0 / S.s_) * self.t_) + self.R_ @ S.t_, self.s_ * S.s_)
+        return Sim2(self.R_ @ S.R_, ((1.0 / S.s_) * self.t_) + self.R_ @ S.t_, self.s_ * S.s_)
```
benjaminrwilson, line 98: Could we consider splitting this up into a few lines? It's tough for me to read all of the operations on one line.

argoverse/utils/sim2.py
```diff
@@ -35,6 +35,24 @@ def __init__(self, R: np.ndarray, t: np.ndarray, s: Union[int, float]) -> None:
...
+        c, s = self.R_[0, 0], self.R_[1, 0]
+        theta_rad = np.arctan2(s, c)
+        return np.rad2deg(theta_rad)
+
+    def __repr__(self) -> str:
+        """Return a human-readable string representation of the class."""
+        return f"Angle (deg.): {self.theta_deg}, Trans.: {np.round(self.t_,2)}, Scale: {self.s_}"
+
     def __eq__(self, other: object) -> bool:
         """Check for equality with other Sim(2) object"""
```
benjaminrwilson, line 57: Nit: period and end of sentence for consistency.

argoverse/utils/sim2.py
```diff
@@ -35,6 +35,24 @@ def __init__(self, R: np.ndarray, t: np.ndarray, s: Union[int, float]) -> None:
...
 
+    @property
+    def theta_deg(self) -> float:
+        """Recover the rotation angle `theta` (in degrees) from the 2d rotation matrix.
+
+        Note: the first column of the rotation matrix R provides sine and cosine of theta,
+            since R is encoded as [c,-s]
+                                  [s, c]
+
+        We use the following identity: tan(theta) = s/c = (opp/hyp) / (adj/hyp) = opp/adj
```
benjaminrwilson, line 46: Nit: period and end of sentence for consistency.

tests/test_sim2.py
```diff
@@ -89,7 +89,37 @@ def test_compose():
...
+
+    # ground truth is an identity transformation
+    aSa_gt = Sim2(R=np.eye(2), t=np.zeros(2), s=1.0)
+    aSa_est = aSb.compose(bSc).compose(aSc.inverse())
+
+    assert np.isclose(aSa_gt.theta_deg, aSa_est.theta_deg, atol=1e-5)
+
+
+def test_inverse() -> None:
     """ """
```
benjaminrwilson, line 123: Could we fill this with something brief?

benjaminrwilson (changes requested):

argoverse/utils/sim2.py
tagarwal-argoai: Missing return statement in the docstring.

argoverse/utils/sim2.py
```diff
@@ -35,6 +35,24 @@ def __init__(self, R: np.ndarray, t: np.ndarray, s: Union[int, float]) -> None:
...
+        """
+        c, s = self.R_[0, 0], self.R_[1, 0]
+        theta_rad = np.arctan2(s, c)
+        return np.rad2deg(theta_rad)
+
+    def __repr__(self) -> str:
+        """Return a human-readable string representation of the class."""
+        return f"Angle (deg.): {self.theta_deg}, Trans.: {np.round(self.t_,2)}, Scale: {self.s_}"
```
tagarwal-argoai, line 54: Can you add the units for `Trans` and `Scale` to be consistent?

argoverse/utils/sim2.py
tagarwal-argoai: +1

tagarwal-argoai (commented):
Left some minor comments.

argoverse/utils/sim2.py
johnwlambert: Thanks for the review, @tagarwal-argoai. So technically, translation (and thus its counterpart scale) are unit-less, as it is up to the user what they choose. It could be inches, meters, centimeters, etc. So I don't think we can include it.

tests/test_sim2.py
johnwlambert: Thanks for the review Ben. Sure, just added that.

tests/test_sim2.py
johnwlambert: thanks, should be updated now

tests/test_sim2.py
johnwlambert: thanks, should be updated now

argoverse/utils/sim2.py
johnwlambert: should be fixed now

argoverse/utils/sim2.py
johnwlambert: Regarding the return statement: I don't think it would add much, except bloat the function.

The Google python style guide says https://google.github.io/styleguide/pyguide.html#383-functions-and-methods:

```
A function must have a docstring, unless it meets all of the following criteria:

not externally visible
very short
obvious
```
I believe these 3 criteria are met here, so I believe it's unnecessary to have the "Returns" section

benjaminrwilson:
Additionally, could we add periods for the rest of the docstrings in `sim2.py`

johnwlambert:
No problem, just added periods to the rest of the docstrings

argoverse/utils/sim2.py
```diff
@@ -68,16 +86,29 @@ def scale(self) -> float:
...
-        return Sim2(self.R_ * S.R_, ((1.0 / S.s_) * self.t_) + self.R_ @ S.t_, self.s_ * S.s_)
+        """Composition with another Sim2.
+
+        This can be understood via block matrix multiplication, if self is parameterized as (R1,t1,s1)
+        and if `S` is parameterized as (R2,t2,s2):
+
+        [R1  t1]   [R2  t2]   [R1 @ R2   R1@t2 + t1/s2]
+        [0 1/s1] @ [0 1/s2] = [ 0          1/(s1*s2)  ]
+        """
+        # fmt: off
```
benjaminrwilson, line 105: Is there a benefit of enabling/disabling the formatting vs. new variables?

benjaminrwilson (changes requested):

argoverse/utils/sim2.py
johnwlambert: @benjaminrwilson so I tried to think of a new variable name that would be clear, but it seemed to just add more complexity and more things to understand.

I think forcing the code onto 3 different lines avoids creating an unnecessary temporary variable and makes it easier to read.

argoverse/utils/sim2.py
tagarwal-argoai: OKay, good to know that we are following some set standards.

argoverse/utils/sim2.py
tagarwal-argoai: I see, thanks for the explanation!

tagarwal-argoai (approved):
Looks good to me!

benjaminrwilson (approved):

PR #244, opened 2021-06-24
Remove quaternion package dependency

johnwlambert:
Switch to fully using [`scipy`](https://github.com/scipy/scipy) (8.3k stars) instead of [`quaternion`](https://github.com/moble/quaternion) (413 stars)

### DIFF
--- a/argoverse/evaluation/competition_util.py
+++ b/argoverse/evaluation/competition_util.py
@@ -10,7 +10,6 @@ from typing import Dict, List, Optional, Tuple, Union
 
 import h5py
 import numpy as np
-import quaternion
 from scipy.spatial import ConvexHull
 from shapely.geometry import Polygon
 from sklearn.cluster.dbscan_ import DBSCAN
@@ -18,6 +17,7 @@ from sklearn.cluster.dbscan_ import DBSCAN
 from argoverse.data_loading.argoverse_tracking_loader import ArgoverseTrackingLoader
 from argoverse.data_loading.object_label_record import ObjectLabelRecord
 from argoverse.utils.se3 import SE3
+from argoverse.utils.transform import rotmat2quat
 
 TYPE_LIST = Union[List[np.ndarray], np.ndarray]
 
@@ -169,6 +169,7 @@ def dist(p1: Tuple[float, float], p2: Tuple[float, float]) -> float:
 
 
 def poly_to_label(poly: Polygon, category: str = "VEHICLE", track_id: str = "") -> ObjectLabelRecord:
+    """ """
     # poly in polygon format
 
     bbox = poly.minimum_rotated_rectangle
@@ -197,18 +198,20 @@ def poly_to_label(poly: Polygon, category: str = "VEHICLE", track_id: str = "") 
     # translation = center
     center = np.array([bbox.centroid.xy[0][0], bbox.centroid.xy[1][0], min(z) + height / 2])
 
+    c = np.cos(angle)
+    s = np.sin(angle)
     R = np.array(
         [
-            [np.cos(angle), -np.sin(angle), 0],
-            [np.sin(angle), np.cos(angle), 0],
+            [c, -s, 0],
+            [s, c, 0],
             [0, 0, 1],
         ]
     )
 
-    q = quaternion.from_rotation_matrix(R)
+    q = rotmat2quat(R)
 
     return ObjectLabelRecord(
-        quaternion=quaternion.as_float_array(q),
+        quaternion=q,
         translation=center,
         length=length,
         width=width,
@@ -266,18 +269,14 @@ def save_label(argoverse_data: ArgoverseTrackingLoader, labels: List[ObjectLabel
     timestamp = argoverse_data.lidar_timestamp_list[idx]
 
     for label in labels:
+        (qw, qx, qy, qz) = label.quaternion
         json_data = {
             "center": {
                 "x": label.translation[0],
                 "y": label.translation[1],
                 "z": label.translation[2],
             },
-            "rotation": {
-                "x": label.quaternion[0],
-                "y": label.quaternion[1],
-                "z": label.quaternion[2],
-                "w": label.quaternion[3],
-            },
+            "rotation": {"x": qx, "y": qy, "z": qz, "w": qw},
             "length": label.length,
             "width": label.width,
             "height": label.height,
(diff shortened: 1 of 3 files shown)


argoverse/evaluation/competition_util.py
```diff
@@ -169,6 +169,7 @@ def dist(p1: Tuple[float, float], p2: Tuple[float, float]) -> float:
 
 
 def poly_to_label(poly: Polygon, category: str = "VEHICLE", track_id: str = "") -> ObjectLabelRecord:
+    """ """
```
benjaminrwilson, line 172: Thanks for taking care of this, @johnwlambert. Could we add a short docstring here?

argoverse/utils/transform.py
```diff
@@ -48,6 +55,13 @@ def quat_argo2scipy(q: np.ndarray) -> np.ndarray:
     return q_scipy
 
 
+def quat_scipy2argo(q: np.ndarray) -> np.ndarray:
```
benjaminrwilson, line 58: This is on a bit of a meta note, but we really need to clean up `transform.py` (the file as a whole).

argoverse/evaluation/competition_util.py
johnwlambert: Sure -- just added one

johnwlambert:
@tagarwal-argoai would you mind taking a look at this, also?

argoverse/evaluation/competition_util.py
```diff
@@ -136,13 +134,9 @@ def get_polygon_from_points(points: np.ndarray) -> Polygon:
...
-    for simplex in hull.simplices:
-        poly.append([points[simplex, 0][0], points[simplex, 1][0], points[simplex, 2][0]])
-        poly.append([points[simplex, 0][1], points[simplex, 1][1], points[simplex, 2][1]])
-
-        # plt.plot(points[simplex, 0], points[simplex, 1], 'k-')
-
-    return Polygon(poly)
+    # `simplices` contains indices of points forming the simplical facets of the convex hull.
+    poly_pts = hull.points[np.unique(hull.simplices)]
+    return Polygon(poly_pts)
```
johnwlambert, line 139: @janesjanes and @benjaminrwilson would you mind taking a brief look at this?

I believe the previous code was actually mixing the x,y,z coordinates of 3 vertices, which I don't think was the intended behavior.

If a simplicial facet is defined by vertex 0 = (x0,y0,z0), vertex 1=(x1,y1,z1), vertex 2=(x2,y2,z2), then I think we were forming (x0,x1,x2) as a point of the polygon, which wouldn't be correct.

argoverse/evaluation/competition_util.py
```diff
@@ -158,57 +152,52 @@ def get_rotated_bbox_from_points(points: np.ndarray) -> Polygon:
     return get_polygon_from_points(points).minimum_rotated_rectangle
 
 
-def unit_vector(pt0: Tuple[float, float], pt1: Tuple[float, float]) -> Tuple[float, float]:
-    # returns an unit vector that points in the direction of pt0 to pt1
-    dis_0_to_1 = math.sqrt((pt0[0] - pt1[0]) ** 2 + (pt0[1] - pt1[1]) ** 2)
-    return (pt1[0] - pt0[0]) / dis_0_to_1, (pt1[1] - pt0[1]) / dis_0_to_1
-
+def poly_to_label(poly: Polygon, category: str = "VEHICLE", track_id: str = "") -> ObjectLabelRecord:
+    """Convert a Shapely Polygon to a 3d cuboid by estimating the minimum-bounding rectangle.
```
johnwlambert, line 156: I tried to simplify this logic a bit here, and make things a bit clearer.

argoverse/evaluation/competition_util.py
```diff
@@ -266,18 +255,14 @@ def save_label(argoverse_data: ArgoverseTrackingLoader, labels: List[ObjectLabel
...
             "center": {
                 "x": label.translation[0],
                 "y": label.translation[1],
                 "z": label.translation[2],
             },
-            "rotation": {
-                "x": label.quaternion[0],
-                "y": label.quaternion[1],
-                "z": label.quaternion[2],
-                "w": label.quaternion[3],
```
johnwlambert, line 279: @janesjanes since we express quaternions as qw, qx, qy, qz, I'm not sure this is correct here.

Since the `quaternion` package uses qw, qx, qy, qz, but we are unpacking this as qx, qy, qz, qw.

argoverse/evaluation/competition_util.py
```diff
@@ -158,57 +152,52 @@ def get_rotated_bbox_from_points(points: np.ndarray) -> Polygon:
...
 
-    x = bbox.exterior.xy[0]
-    y = bbox.exterior.xy[1]
     z = np.array([z for _, _, z in poly.exterior.coords])
+    height = max(z) - min(z)
 
-    # z = poly.exterior.xy[2]
-
-    d1 = dist((x[0], y[0]), (x[1], y[1]))
-    d2 = dist((x[1], y[1]), (x[2], y[2]))
```
johnwlambert, line 183: i thought this indexing was a bit hard to follow, so i tried to simplify it a bit

argoverse/evaluation/competition_util.py
benjaminrwilson: Thanks for looking into this, @johnwlambert. Before I dive in, are we using this function anywhere?

argoverse/evaluation/competition_util.py
benjaminrwilson: Does this get called outside of this file?

argoverse/evaluation/competition_util.py
johnwlambert: Thanks Ben. So this is used in a tracking competition demo notebook: https://github.com/argoai/argoverse-api/blob/master/demo_usage/competition_tracking_tutorial.ipynb

argoverse/evaluation/competition_util.py
johnwlambert: Just here: https://github.com/argoai/argoverse-api/blob/master/demo_usage/competition_tracking_tutorial.ipynb

argoverse/evaluation/competition_util.py
```diff
@@ -220,7 +206,7 @@ def poly_to_label(poly: Polygon, category: str = "VEHICLE", track_id: str = "")
 
 
 def get_objects(clustering: DBSCAN, pts: np.ndarray, category: str = "VEHICLE") -> List[Tuple[np.ndarray, uuid.UUID]]:
-
+    """ """
```
benjaminrwilson, line 209: Could we provide a small docstring here?

PR #245, opened 2021-06-24
Use only one significant fig in Sim2 __repr__

johnwlambert:
Prevent needlessly verbose messages from too many sig figs, when printing with the __ repr __ method


benjaminrwilson:
LGTM 👍

PR #246, opened 2021-06-27
Bump hydra + omegaconf.

benjaminrwilson:
Compatible releases of `hydra` and `omegaconf` are available.


johnwlambert:
Thanks for the PR, Ben. Good we dont have to use the dev prerelease versions anymore.

My understanding is that if we don't pin down the new version number, then if someone went to install argoverse-api, and had an old version of hydra-core on their system, then those dependencies would not be updated, and could lead to missing functionality.

benjaminrwilson:
In that case, do you think we should pin all of the dependencies?

johnwlambert:
I think we've already pinned down the ones where we saw people having issues, or where we saw some significant changes between versions. Maybe we can just pin down hydra and omegaconf for now?

setup.py
```diff
@@ -47,18 +47,18 @@
     packages=find_packages(exclude=["tests", "integration_tests", "map_files"]),
     package_data={"argoverse": ["argoverse/visualization/data/**/*"]},
     include_package_data=True,
-    python_requires=">= 3.6",
+    python_requires=">= 3.7",
```
benjaminrwilson, line 50: Since 3.6 has been deprecated.

johnwlambert (approved):

PR #248, opened 2021-06-28
add rotation matrix -> quaternion and yaw -> quaternion conversion utility functions

johnwlambert:
Add rotation matrix -> quaternion and yaw -> quaternion conversion utility functions.

Previously, users used additional external libraries like `quaternion` for this functionality.


benjaminrwilson:
LGTM. Thanks, @johnwlambert!

PR #252, opened 2021-07-01
Create parent dir, when serializing Sim(2), if it doesnt exist

johnwlambert:
Instead of crashing, simply create parent dir of requested absolute filepath.

### DIFF
--- a/argoverse/utils/sim2.py
+++ b/argoverse/utils/sim2.py
@@ -8,6 +8,7 @@ References:
 
 import json
 import os
+from pathlib import Path
 from typing import Union
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


argoverse/utils/sim2.py
```diff
@@ -156,7 +156,7 @@ def save_as_json(self, save_fpath: _PathLike) -> None:
         """
         if not Path(save_fpath).exists():
```
benjaminrwilson, line 157: Hi @jlambert,

Thanks for adding this. I notice that this function calls `save_json_dict`. Do you think that it might be convenient if we move directory creation into that function?

My thought is this: https://github.com/argoai/argoverse-api/blob/9ad7251e863590603c3f6064a45830dabd3a15e2/argoverse/utils/json_utils.py#L32 could be modified as:

```
json_fpath = Path(json_fpath)
json_fpath.parent.mkdir(parents=True, exist_ok=True)
with json_fpath.open("w") as f:
    json.dump(dictionary, f)

```

`Path(save_fpath).parent.mkdir(parents=True, exists_ok=True)`

tagarwal-argoai: +1 on this.

tagarwal-argoai (commented):
Left a small comment.

benjaminrwilson:
@johnwlambert, is there anything I can do to help with this PR?

PR #254, opened 2021-07-07
Updated workflow + removed windows check.

benjaminrwilson:
- Updated Github Actions to include MacOS and Windows.
- Removed explicit system exit on Windows.


.github/workflows/test-python.yml
```diff
@@ -7,18 +7,30 @@ jobs:
...
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python 3.7
+    - name: Set up Python ${{ matrix.python }}
       uses: actions/setup-python@v2
       with:
-        python-version: 3.7.1
+        python-version: ${{ matrix.python }}
+
+    # - if: matrix.platform == 'windows-latest'
```
johnwlambert, line 23: if we uncomment this, does it crash?

.github/workflows/test-python.yml
```diff
@@ -7,18 +7,30 @@ jobs:
...
+    #   run: |
+    #     choco install wget --no-progress
+
+    - if: matrix.platform == 'macos-latest'
+      run: |
+        brew install opencv@2
```
johnwlambert, line 29: would prefer to use a miniconda CI rather than brew: https://github.com/borglab/gtsfm/blob/master/.github/workflows/test-python.yml

.github/workflows/test-python.yml
```diff
@@ -7,18 +7,30 @@ jobs:
...
 
     - name: Install dependencies
       run: |
-        pip install tox==3.12.1 pre-commit==1.17.0
+        pip install tox pre-commit
```
johnwlambert, line 33: any reason to untag the version?

setup.py
```diff
@@ -20,11 +20,6 @@
 with open(path.join(here, "README.md"), encoding="utf-8") as f:
     long_description = f.read()
 
-
-if platform.system() == "Windows":
```
johnwlambert, line 24: we should only remove this once the Windows CI passes i think

.github/workflows/test-python.yml
benjaminrwilson: No. I added this for windows support. We can put it back in -- we'd definitely need it to get windows working.

pyproject.toml
```diff
@@ -1,3 +1,7 @@
+[build-system]
+requires = ["setuptools", "wheel", "numpy"]
+build-backend = "setuptools.build_meta"
+
```
benjaminrwilson, line 4: https://setuptools.readthedocs.io/en/latest/userguide/quickstart.html#python-packaging-at-a-glance

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,52 @@
...
+      - uses: actions/download-artifact@v2
+        with:
+          name: artifact
+          path: dist
+
+      - uses: pypa/gh-action-pypi-publish@v1.4.2
+        with:
+          user: __token__
+          password: ${{ secrets.pypi_password }}
+          # To test: repository_url: https://test.pypi.org/legacy/
```
benjaminrwilson, line 52: https://github.com/pypa/cibuildwheel/blob/main/examples/github-deploy.yml

pyproject.toml
```diff
@@ -1,3 +1,7 @@
+[build-system]
+requires = ["setuptools", "wheel", "numpy"]
```
benjaminrwilson, line 2: `numpy` is a build dependency for `lap`.

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,48 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Build wheels and sdist on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
```
johnwlambert, line 11: nice work, Ben. so we don't have any compatibility issues with windows anymore? i thought the "/path/dirpath/fname" hardcoding was causing issues there?

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,48 @@
+name: Build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Build wheels and sdist on ${{ matrix.os }}
```
johnwlambert, line 7: could you maybe lengthen the name a bit to explain what sdist is? or put that in a comment?

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,48 @@
...
+          path: |
+            dist/*.whl
+            dist/*.tar.gz
+
+# Disable until PyPI integration.
+  # upload_pypi:
+  #   needs: [build]
+  #   runs-on: ${{ matrix.os }}
+  #   # upload to PyPI on every tag starting with 'v'
+  #   if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
```
johnwlambert, line 38: would we want to push to pypi on every push? i would think only on a merge or opening a pull request, right?

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,48 @@
+name: Build
+
```
johnwlambert, line 2: is there any workflow or file we are basing this off of? if so, should we add a reference to it?

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,48 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Build wheels
+        run: |
+          pip install -U build
```
johnwlambert, line 24: what does the -U do? I thought we needed twine for this, but is that only if we're compiling something?

.github/workflows/build_wheels.yml
johnwlambert: i guess before we build the windows wheel, i think we should make sure the tests pass on windows first

so we dont give people false hope at this stage.

.github/workflows/build_wheels.yml
johnwlambert: i guess twine is just needed for the PyPI push

.github/workflows/build_wheels.yml
benjaminrwilson: I agree that wouldn't be good -- the second condition ensures that it checks for things prefixed with "v" like the a "v1.0" release (this was taken straight from the `cibuildwheel` repository: https://github.com/pypa/cibuildwheel/blob/c936190712ab49c83cf113afe3ca0f63f02aa9a4/examples/github-deploy.yml#L58).

.github/workflows/build_wheels.yml
benjaminrwilson: Thanks, John. I think that could be causing issues, but I think it would probably show up in the tests. I'll work on modifying our test workflow accordingly.

benjaminrwilson: Made some modifications to the testing workflow. We're blocked by `disparity_interpolation`, but @jhonykaesemodel and I have discussed a longer term fix to simplify the build process.

.github/workflows/build_wheels.yml
benjaminrwilson: `-U, --upgrade               Upgrade all specified packages to the newest available version. The handling of dependencies
                              depends on the upgrade-strategy used.`

PR #257, opened 2021-07-22
Pin `numpy` to 1.2.1

benjaminrwilson:


argoverse/data_loading/argoverse_tracking_loader.py
```diff
@@ -25,7 +26,7 @@ def __init__(self, root_dir: str) -> None:
         self.CAMERA_LIST = CAMERA_LIST
         self._log_list: Optional[List[str]] = None
         self._image_list: Optional[Dict[str, Dict[str, List[str]]]] = None
-        self._image_list_sync: Optional[Dict[str, Dict[str, List[np.ndarray]]]] = None
+        self._image_list_sync: Optional[Dict[str, Dict[str, List[NDArray[np.float64]]]]] = None
```
johnwlambert, line 29: Hi Anna, the image timestamps should be in nanoseconds (integers), so likely we need to cast to np.int32 or np.int64 somewhere beforehand

argoverse/data_loading/argoverse_forecasting_loader.py
```diff
@@ -77,7 +77,7 @@ def seq_df(self) -> pd.DataFrame:
         return _read_csv(self.current_seq)
 
     @property
-    def agent_traj(self) -> np.ndarray:
+    def agent_traj(self) -> Any:
```
johnwlambert, line 80: maybe we can cast the return arg to np.array() since, np.column_stack() should hopefully only return an array

argoverse/evaluation/detection/eval.py
```diff
@@ -117,7 +118,7 @@ def evaluate(self) -> pd.DataFrame:
         gt_fpaths = list(self.gt_root_fpath.glob("*/per_sweep_annotations_amodal/*.json"))
 
         assert len(dt_fpaths) == len(gt_fpaths)
-        data: DefaultDict[str, List[np.ndarray]] = defaultdict(list)
+        data: DefaultDict[str, List[NDArray[np.float64]]] = defaultdict(list)
```
johnwlambert, line 121: Hi @ahuller3, thanks for your work on this. 

@benjaminrwilson and I were reviewing this today, and we were thinking it might be easier for readability to do something like

if we create something like `argoverse/typing/numpy.py`
```python
NDArrayFloat = Union[NDArray[np.float32], NDArray[np.float64]]
```

What are your thoughts?

PR #258, opened 2021-07-23
Update for deprecated types.

benjaminrwilson:
This maps `np.int` -> `int` and `np.float` -> `float` as the former values are deprecated aliases of the latter.


johnwlambert (approved):

johnwlambert:
Deprecated in Numpy 1.20.0: https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

PR #259, opened 2021-07-23
Migrate to setup.cfg.

benjaminrwilson:
https://setuptools.readthedocs.io/en/latest/userguide/quickstart.html#transitioning-from-setup-py-to-setup-cfg

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,57 @@
 [metadata]
+author = Argo AI
+author_email = argoverse-api@argo.ai
+classifiers = 
+    Development Status :: 5 - Production/Stable
+    Intended Audience :: Developers
+    License :: OSI Approved :: MIT License
+    Operating System :: POSIX
+    Operating System :: MacOS
+    Programming Language :: Python :: 3
+    Topic :: Scientific/Engineering :: Artificial Intelligence
 description_file = README.md
+keywords =
+    self-driving-car
+    dataset-tools
+license = MIT
+name = argoverse
+url = https://www.argoverse.org
+version = attr: argoverse.VERSION
+
+[options]
+zip_safe = False
+include_package_data = True
+packages = find:
+scripts =
+    bin/first.py
+    bin/second.py
+install_requires =
+    colour
+    descartes
+    imageio
+    h5py
+    hydra-core ==1.1.0
+    lap
+    matplotlib
+    motmetrics ==1.1.3
+    numpy ==1.19
+    omegaconf ==2.1.0
+    opencv-python >=4.1.0.25
+    pandas >=0.23.1
+    pillow
+    imageio
+    pyntcloud >=0.1.0
+    scipy >=1.4.0
+    shapely
+    sklearn
+    tqdm
+    typing_extensions
+
+[options.package_data]
+argoverse = argoverse/visualization/data/**/*
+
+[options.packages.find]
+exclude =
+    integration_tests
+    map_files
+    tests
--- a/setup.py
+++ b/setup.py
@@ -7,67 +7,12 @@ https://github.com/pypa/sampleproject
 """
 
 import platform
-import sys
-from codecs import open  # To use a consistent encoding
-from os import path
 
 # Always prefer setuptools over distutils
 from setuptools import find_packages, setup
 
-here = path.abspath(path.dirname(__file__))
-
-# Get the long description from the README file
-with open(path.join(here, "README.md"), encoding="utf-8") as f:
-    long_description = f.read()
-
-
 if platform.system() == "Windows":
     print("Argoverse currently does not support Windows, please use Linux/Mac OS")
-    sys.exit(1)
+    exit(1)
 
-setup(
-    name="argoverse",
-    version="1.0.1",
-    description="",
-    long_description=long_description,
-    url="https://www.argoverse.org",
-    author="Argo AI",
-    author_email="argoverse-api@argo.ai",
-    license="MIT",
-    classifiers=[
-        "Development Status :: 5 - Production/Stable",
-        "Intended Audience :: Developers",
-        "License :: OSI Approved :: MIT License",
-        "Operating System :: POSIX",
-        "Operating System :: MacOS",
-        "Programming Language :: Python :: 3",
-        "Topic :: Scientific/Engineering :: Artificial Intelligence",
-    ],
-    keywords="self-driving-car dataset-tools",
-    packages=find_packages(exclude=["tests", "integration_tests", "map_files"]),
-    package_data={"argoverse": ["argoverse/visualization/data/**/*"]},
-    include_package_data=True,
-    python_requires=">= 3.7",
-    install_requires=[
-        "colour",
-        "descartes",
-        "imageio",
-        "h5py",
-        "hydra-core==1.1.0",
-        "lap",
-        "matplotlib",
-        "motmetrics==1.1.3",
-        "numpy==1.19",
-        "omegaconf==2.1.0",
-        "opencv-python>=4.1.0.25",
-        "pandas>=0.23.1",
-        "pillow",
-        "imageio",
-        "pyntcloud>=0.1.0",
-        "scipy>=1.4.0",
-        "shapely",
-        "sklearn",
-        "tqdm",
-        "typing_extensions",
-    ],
-)
+setup()


demo_usage/visualize_30hz_benchmark_data_on_map.py
```diff
@@ -247,7 +247,7 @@ def render_bev_labels_mpl(
             city_to_egovehicle_se3: Transformation from egovehicle frame to city frame
             avm: ArgoverseMap instance
         """
-        if axis is not "city_axis":
+        if axis != "city_axis":
```
benjaminrwilson, line 250: Flagged by `flake8`.

demo_usage/visualize_30hz_benchmark_data_on_map.py
```diff
@@ -262,7 +262,7 @@ def render_bev_labels_mpl(
         draw_lane_polygons(ax, local_lane_polygons)
         draw_lane_polygons(ax, local_das, color="tab:pink")
 
-        if axis is not "city_axis":
+        if axis != "city_axis":
```
benjaminrwilson, line 265: Flagged by `flake8`.

demo_usage/visualize_30hz_benchmark_data_on_map.py
```diff
@@ -283,7 +283,7 @@ def render_bev_labels_mpl(
                     bbox_ego_frame = rotate_polygon_about_pt(
                         bbox_ego_frame, city_to_egovehicle_se3.rotation, np.zeros((3,))
                     )
-                    if axis is "city_axis":
+                    if axis == "city_axis":
```
benjaminrwilson, line 286: Flagged by `flake8`.

pyproject.toml
```diff
@@ -6,5 +9,9 @@ line_length = 120
 multi_line_output = 3
 profile = "black"
 
-[tool.black]
-line-length = 120
+[tool.mypy]
+ignore_missing_imports = true
+strict = true
```
benjaminrwilson, line 14: `mypy` settings are supported in `pyproject.toml` now.

pyproject.toml
```diff
@@ -6,5 +9,9 @@ line_length = 120
...
+
+[tool.pytest.ini_options]
+addopts = "--cov argoverse --cov-append --cov-branch --cov-report=term-missing"
```
benjaminrwilson, line 17: Add `pytest` settings into `pyproject.toml`.

setup.cfg
```diff
@@ -1,2 +1,59 @@
...
+[options.packages.find]
+exclude =
+    integration_tests
+    map_files
+    tests
+
+[flake8]
+enable_extensions = G,argoverse
+ignore = E203,E704,E711,E722,E741,W291,W293,W391,W503,F821,F401,F811,F841,P101,G004,G002,I201,I100,I101
+max_line_length = 120
```
benjaminrwilson, line 59: `flake8` configuration in `setup.cfg` -- does not support `pyproject.toml` currently.

tox.ini
```diff
@@ -13,9 +13,10 @@ deps =
 
 commands =
    python3 -m pip install numba
-   pytest tests --cov argoverse --cov-append --cov-branch --cov-report=term-missing
-   flake8 --max-line-length 120 --ignore E203,E704,E711,E722,E741,W291,W293,W391,W503,F821,F401,F811,F841,P101,G004,G002,I201,I100,I101 --enable-extensions G argoverse
-   mypy --ignore-missing --strict argoverse
+   pytest tests
+   flake8
+   mypy
+
```
benjaminrwilson, line 19: These now reside in the global configuration.

demo_usage/cuboids_to_bboxes.py
```diff
@@ -348,7 +348,7 @@ def main(args: Any):
     logger.info(args)
 
     if args.log_ids is None:
-        logger.error(f"Please provide a comma seperated list of log ids")
-        raise ValueError(f"Please provide a comma seperated list of log ids")
+        logger.error("Please provide a comma seperated list of log ids")
+        raise ValueError("Please provide a comma seperated list of log ids")
```
benjaminrwilson, line 352: Flagged by `flake8`.

argoverse/__init__.py
```diff
@@ -0,0 +1 @@
+VERSION = "1.1.0"
```
benjaminrwilson, line 1: Used in `setup.cfg`.

pyproject.toml
```diff
@@ -1,3 +1,6 @@
+[tool.black]
+line-length = 120
+
```
benjaminrwilson, line 3: Sorted alphabetically.

setup.cfg
```diff
@@ -1,2 +1,59 @@
 [metadata]
```
benjaminrwilson, line 1: Static declaration of metadata previously listed in `setup.py`.

tox.ini
```diff
@@ -13,9 +13,10 @@ deps =
 
 commands =
    python3 -m pip install numba
-   pytest tests --cov argoverse --cov-append --cov-branch --cov-report=term-missing
-   flake8 --max-line-length 120 --ignore E203,E704,E711,E722,E741,W291,W293,W391,W503,F821,F401,F811,F841,P101,G004,G002,I201,I100,I101 --enable-extensions G argoverse
-   mypy --ignore-missing --strict argoverse
+   pytest tests
+   flake8 argoverse
+   mypy argoverse
+
```
benjaminrwilson, line 19: These flags reside in either `setup.cfg` or `pyproject.toml` now.

wqi (approved):
Changes LGTM

johnwlambert:
Thanks for the PR, Ben. Could you add to the PR description a few more details about the advantages of making this change?

johnwlambert:
Could you add the version numbers as well, also? Do we need a specific version of `setuptools` now to install this? And for any python version?

demo_usage/cuboids_to_bboxes.py
johnwlambert: nice catch, thanks

argoverse/__init__.py
johnwlambert: is the version required to be placed here? This probably wouldnt be a place I would look for the version number, so do we need to write it in 2 places now?

demo_usage/visualize_30hz_benchmark_data_on_map.py
johnwlambert: thanks

demo_usage/visualize_30hz_benchmark_data_on_map.py
johnwlambert: thanks

demo_usage/visualize_30hz_benchmark_data_on_map.py
johnwlambert: thanks

pyproject.toml
johnwlambert: is there a specific version of mypy that we need to enforce, to be able to use this functionality?

setup.cfg
```diff
@@ -1,2 +1,60 @@
...
+keywords =
+    self-driving-car
+    dataset-tools
+license = MIT
+name = argoverse
+url = https://www.argoverse.org
+version = attr: argoverse.VERSION
+
+[options]
+zip_safe = False
```
johnwlambert, line 22: why do we want zip_safe = False?

tox.ini
```diff
@@ -12,9 +12,9 @@ deps =
     flake8-string-format
 
 commands =
-   pytest tests --cov argoverse --cov-append --cov-branch --cov-report=term-missing
-   flake8 --max-line-length 120 --ignore E203,E704,E711,E722,E741,W291,W293,W391,W503,F821,F401,F811,F841,P101,G004,G002,I201,I100,I101 --enable-extensions G argoverse
-   mypy --ignore-missing --strict argoverse
+   pytest tests
```
johnwlambert, line 15: have we verified that these options get picked up from the pyproject.toml file in practice?

setup.cfg
benjaminrwilson: This was just a default option listed here: https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html?highlight=zip_safe. I don't have a strong opinion on it. `setuptools` will automatically try to determine the setting if we remove it, which might be useful.

argoverse/__init__.py
benjaminrwilson: This is the first method listed here: https://packaging.python.org/guides/single-sourcing-package-version/#single-sourcing-the-package-version. Although, they use `__version__` instead of `VERSION`.

benjaminrwilson:
> Could you add the version numbers as well, also?

Which version numbers are you referring to?

>  Do we need a specific version of `setuptools` now to install this? And for any python version?

`New in 30.3.0 (8 Dec 2016).`

https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html#configuring-setup-using-setup-cfg-files

argoverse/__init__.py
benjaminrwilson: Updated to use `__version__`.

tox.ini
benjaminrwilson: Some of these can be verified by looking at the `tox` run in our Github workflow. I'll verify once again after we've settled on the other changes.

pyproject.toml
benjaminrwilson: Do we use `mypy` anywhere other than: https://github.com/argoai/argoverse-api/blob/ceaee0f0e09281fc36f58900388a0d00fd6841f9/tox.ini?

tagarwal-argoai (approved):
I took a look at the PR and it looks good overall. I am not an expert with setup.cfg but I trust your and John’s review on this and you could proceed with landing this.

PR #275, opened 2021-10-12
Synchronize version + add `polars`

benjaminrwilson:
Our versioning appears to be out of date with our most recent release (1.0.1 vs. 1.1.0). Additionally, adding `polars` as a dependency for AV2 development + alphabetizing list.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ if platform.system() == "Windows":
 
 setup(
     name="argoverse",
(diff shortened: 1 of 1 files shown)


johnwlambert:
Thanks for the PR @benjaminrwilson. Looks good.

Polars is the rust-implemented dataframes lib? What's the speedup over pandas with it for your use cases?

PR #285, opened 2021-11-16
Add map root path when visualizing data

purewater0901:
Hello argoverse-api developers. I add a new argument to the function viz_sequence so that we can indicate the map root path when visualizing map and trajectory data. My change might not follow coding rules of your team. If so, I am happy to change my code to follow your guidelines. Thanks.

### DIFF
--- a/argoverse/visualization/visualize_sequences.py
+++ b/argoverse/visualization/visualize_sequences.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
 
 """A simple python script template."""
+import os
 from collections import defaultdict
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Union
 
 import matplotlib.lines as mlines
 import matplotlib.pyplot as plt
@@ -14,6 +15,7 @@ from argoverse.map_representation.map_api import ArgoverseMap
 
 _ZORDER = {"AGENT": 15, "AV": 10, "OTHERS": 5}
 
+_PathLike = Union[str, "os.PathLike[str]", None]
 
 def interpolate_polyline(polyline: np.ndarray, num_points: int) -> np.ndarray:
     duplicates = []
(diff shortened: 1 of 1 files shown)


argoverse/visualization/visualize_sequences.py
```diff
@@ -14,6 +15,7 @@
 
 _ZORDER = {"AGENT": 15, "AV": 10, "OTHERS": 5}
 
+_PathLike = Union[str, "os.PathLike[str]", None]
```
mgilson-argo, line 18: I think this would be more natural as:

```py
_PathLike = Union[str, "os.PathLike[str]"]
```

Then it is up to a user to use an `Optional[_PathLike]` if the `None` is permissable.

argoverse/visualization/visualize_sequences.py
```diff
@@ -41,7 +44,10 @@ def viz_sequence(
 
     if lane_centerlines is None:
         # Get API for Argo Dataset map
-        avm = ArgoverseMap()
+        if map_root is not None:
+            avm = ArgoverseMap(map_root)
```
mgilson-argo, line 48: What's the default in `ArgoverseMap` if `map_root` isn't passed?  I would have expected it to just be `None` anyway.  If that's the case, you could just do:

```py
avm = ArgoverseMap(map_root)
```

Either way...

argoverse/visualization/visualize_sequences.py
purewater0901: @mgilson-argo Thank you for your review, and I modified the PathLike expression based on your advice.

argoverse/visualization/visualize_sequences.py
purewater0901: If you do not pass the `map_root` to `ArogverseMap`, it would be `ROOT=Path(__file__).resolve().parent.parent.parent / "map_files"`. 

Related Links:
- https://github.com/argoai/argoverse-api/blob/master/argoverse/map_representation/map_api.py#L41
- https://github.com/argoai/argoverse-api/blob/master/argoverse/map_representation/map_api.py#L58

PR #292, opened 2022-01-25
Fix map api docstrings.

johnwlambert:


johnwlambert:
@benjaminrwilson would you mind taking a look at this when you get a chance?

argoverse/map_representation/map_api.py
```diff
@@ -337,9 +336,7 @@ def remove_ground_surface(
         return point_cloud[not_ground_indxs]
 
     def remove_non_driveable_area_points(self, point_cloud: np.ndarray, city_name: str) -> np.ndarray:
-        """
-        Get a lidar point, snap it to the grid, perform the O(1) raster map query.
-        If our z-height is within THRESHOLD of that grid's z-height, then we keep it; otherwise, discard it
```
tagarwal-argoai, line 342: [nit]: Don't we need this line anymore?

johnwlambert: this line was erroneously copied from the function above; ground height is not considered here when determining points that fall within the drivable area

tagarwal-argoai (approved):

